PR #31, opened 2015-11-27
Fix failure writing UUID, URL and URI

komamitsu:
I found that writing UUID, URL and URI as a Map value fails.


cowtowncoder:
Good catch, thank you!

komamitsu:
@cowtowncoder Thanks!
BTW, when will `2.7.0-rc2` be released?

cowtowncoder:
@komamitsu I just pushed 2.7.0-rc1 out (well, almost complete over past couple of days), so probably it will be next weekend until rc2.

komamitsu:
I see. Thanks

Issue #55, opened 2017-08-14
ArrayIndexOutOfBounds using ServiceLoader with java 9


brharrington:
I'm not sure if this is a jackson bug or a java bug, so this might just be an FYI if you haven't encountered it already. So far the jackson-jr-objects jar is the only one that I know causing this problem. Is there anything special about how this jar is created?

Simple test program below, note it doesn't actually use jackson at all but having jackson on the classpath can cause the problem due to the service loader scanning.

```java
import java.util.Iterator;
import java.util.ServiceLoader;

public class Test {
  public static void main(String[] args) throws Exception {
    ServiceLoader<String> loader = ServiceLoader.load(String.class);
    Iterator<String> iter = loader.iterator();
    while (iter.hasNext()) {
      System.out.println(iter.next());
    }
  }
}
```

Results on java 9:

```
$ java -version
java version "9"
Java(TM) SE Runtime Environment (build 9+180)
Java HotSpot(TM) 64-Bit Server VM (build 9+180, mixed mode)
$ java Test
$ java -classpath .:jackson-core-2.9.0.jar Test
$ java -classpath .:jackson-core-2.9.0.jar:jackson-jr-objects-2.9.0.jar Test
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 64
	at java.base/java.util.jar.JarFile.match(JarFile.java:941)
	at java.base/java.util.jar.JarFile.checkForSpecialAttributes(JarFile.java:971)
	at java.base/java.util.jar.JarFile.isMultiRelease(JarFile.java:366)
	at java.base/java.util.jar.JarFile.getEntry(JarFile.java:491)
	at java.base/java.util.jar.JarFile.getJarEntry(JarFile.java:447)
	at java.base/jdk.internal.util.jar.JarIndex.getJarIndex(JarIndex.java:115)
	at java.base/jdk.internal.loader.URLClassPath$JarLoader$1.run(URLClassPath.java:692)
	at java.base/jdk.internal.loader.URLClassPath$JarLoader$1.run(URLClassPath.java:684)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/jdk.internal.loader.URLClassPath$JarLoader.ensureOpen(URLClassPath.java:683)
	at java.base/jdk.internal.loader.URLClassPath$JarLoader.<init>(URLClassPath.java:658)
	at java.base/jdk.internal.loader.URLClassPath$3.run(URLClassPath.java:426)
[...]

PR #56, opened 2017-08-14
avoid using back ticks in jar manifest

brharrington:
Current versions of jdk9 have a bug in the JarFile.match
function which throws an ArrayIndexOutOfBoundsException
if a back tick is encountered. This change works around
the issue by removing the back ticks from the build
description. See #55 for test case.

### DIFF
--- a/jr-objects/pom.xml
+++ b/jr-objects/pom.xml
@@ -7,7 +7,7 @@
   </parent>
   <artifactId>jackson-jr-objects</artifactId>
   <packaging>bundle</packaging>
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Interesting. Ideally it'd be good not to have such limitations, but since backticks are there just because of inclusion in wiki -- or more specifically, since formatting matters for github project, not for pom -- I think this makes sense.

brharrington:
Thanks. Still waiting to hear back from Oracle, but if they accept the bug report I'll link the JDK issue here for reference.

cowtowncoder:
Thank you for following up with Oracle too.

PR #85, opened 2021-05-13
Implement equals for Jrs types

MikeDombo:
Implementing equals and hashCode for all the Jrs* types which were missing it. I simply used IntelliJ's automatic generator to make the code.


cowtowncoder:
Sounds good: I'll merge this and then add a unit test or two I think.

Issue #91, opened 2022-01-04
Annotation support should allow `@JsonValue`/`JsonCreator` on `enum`


ashleyfrieze:
As suggested in https://cowtowncoder.medium.com/using-jackson-annotations-with-jackson-jr-51087850b95e this would be an important addition to jackson jr.

Our context is that we're using jackson jr in an AWS Lambda in order to reduce code size. We have an open api module which has code-generated Java classes/enums, and we have the following generated code:

```java
public enum VehicleType {
    CAR("Car"),
    MOTORCYCLE("Motorcycle"),
    ...


    @JsonValue
    public String toString() {
        return String.valueOf(this.value);
    }

    @JsonCreator
    public static VehicleType fromValue(String value) {
```

As things stand at the moment, there's no way we can deserialize `Car` into `VehicleType.CAR`

ashleyfrieze:
I've hacked together a solution which will bridge the gap for us until a proper fix is available:

```java
public class EnumCompatibleAnnotationExtension extends JacksonJrExtension {

    @Override
    protected void register(ExtensionContext ctxt) {
        ctxt.insertModifier(new EnumCompatibleAnnotationModifier(JsonAutoDetect.Value.defaultVisibility()));
    }

    private static class EnumCompatibleAnnotationModifier extends AnnotationBasedValueRWModifier {

        public EnumCompatibleAnnotationModifier(JsonAutoDetect.Value visibility) {
            super(visibility);
        }

        @Override
        public ValueWriter overrideStandardValueWriter(JSONWriter writeContext, Class<?> type, int stdTypeId) {
            if (stdTypeId == SER_ENUM_ID) {
                return new EnumReaderWriter(type);
            }
            return null;
        }

        @Override
        public ValueReader modifyValueReader(JSONReader readContext, Class<?> type, ValueReader defaultReader) {
            if (type.isEnum()) {
                return new EnumReaderWriter(type);
            }
            return super.modifyValueReader(readContext, type, defaultReader);
        }
    }

    @FunctionalInterface
    private interface ConversionFunction<T, R> {
        R apply(T value) throws IOException;
    }

    private static class EnumReaderWriter extends ValueReader implements ValueWriter {
        private final Map<String, String> enumMap;
        private final ConversionFunction<Object, String> writer;
        private final ConversionFunction<String, Object> reader;

        public EnumReaderWriter(Class<?> type) {
            super(type);
            enumMap = new HashMap<>();
[...]

PR #92, opened 2022-01-05
Support `@JsonValue` and `@JsonCreator` for `enum`

ashleyfrieze:
This is a potential resolution for #91 - it's based on the snippet posted, but has been re-written to better follow the house style, and to avoid balling up the whole implementation into a single class.

It comes with tests and updates to the `README.md`

### DIFF
--- a/jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
+++ b/jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
@@ -2,19 +2,28 @@ package com.fasterxml.jackson.jr.annotationsupport;
 
 import java.io.IOException;
 import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.util.*;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import com.fasterxml.jackson.annotation.JsonValue;
 import com.fasterxml.jackson.core.JsonGenerator;
 
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.jr.ob.JSONObjectException;
 import com.fasterxml.jackson.jr.ob.api.ReaderWriterModifier;
+import com.fasterxml.jackson.jr.ob.api.ValueReader;
 import com.fasterxml.jackson.jr.ob.api.ValueWriter;
 import com.fasterxml.jackson.jr.ob.impl.JSONReader;
 import com.fasterxml.jackson.jr.ob.impl.JSONWriter;
 import com.fasterxml.jackson.jr.ob.impl.POJODefinition;
 import com.fasterxml.jackson.jr.ob.impl.ValueLocatorBase;
+
+import static com.fasterxml.jackson.jr.ob.impl.Types.isEnum;
 
 public class AnnotationBasedValueRWModifier extends ReaderWriterModifier
 {
@@ -43,9 +52,27 @@ public class AnnotationBasedValueRWModifier extends ReaderWriterModifier
     @Override
     public ValueWriter overrideStandardValueWriter(JSONWriter writeContext, Class<?> type, int stdTypeId) {
         if (stdTypeId == SER_ENUM_ID) {
-            return new EnumWriter(type);
+            ValueWriter writeByJsonValue = EnumJsonValueWriter.of(type);
+            if (writeByJsonValue != null) {
+                return writeByJsonValue;
+            }
+            // the EnumWriter requires this to be the actual enum type, rather than a subclass of Enum
+            if (type.isEnum()) {
+                return new EnumWriter(type);
+            }
         }
         return null;
+    }
+
+    @Override
+    public ValueReader modifyValueReader(JSONReader readContext, Class<?> type, ValueReader defaultReader) {
+        if (type.isEnum()) {
+            ValueReader readUsingJsonCreator = EnumJsonCreatorReader.of(type);
+            if (readUsingJsonCreator != null) {
+                return readUsingJsonCreator;
+            }
+        }
+        return super.modifyValueReader(readContext, type, defaultReader);
     }
 
     private static class EnumWriter implements ValueWriter {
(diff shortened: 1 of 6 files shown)


ashleyfrieze:
@cowtowncoder I'd appreciate your thoughts on this

cowtowncoder:
Looks good, so far.

One thing to try, I think, is to try bit more special Enums; one that creates sub-classes -- ones where there are one or more methods overridden for individual Enum entries (let me know if that is unclear, I'll add/find an example). The reason is that this case produces one more level of Enum implementation classes, and often causes issues for frameworks. Jackson handles these fine and it's not super hard to do, but requires some care.

ashleyfrieze:
@cowtowncoder if you can find or invent an example, that would be great. My expectation is that this should just work because of the `@JsonCreator` static function.

cowtowncoder:
The problem is that while `static`, it is not declared in actual Enum implementation (sub-)class, but on its parent class.
So if Reflection looks at sub-class it will not see that method.

From `EnumCreatorTest` of jackson-databind we have:

```
    protected enum TestEnumFromInt
    {
        ENUM_A(1), ENUM_B(2), ENUM_C(3);
        
        private final int id;
        
        private TestEnumFromInt(int id) {
            this.id = id;
        }
        
        @JsonCreator public static TestEnumFromInt fromId(int id) {
            for (TestEnumFromInt e: values()) {
                if (e.id == id) return e;
            }
            return null;
        }
    }
```

which kind of shows the case I am worried about. Sub-classes are (and must be) generated in such cases; either for non-default constructor or possibly override of methods. Latter would look like...

```
    protected enum AnotherEnum
    {
        ENUM_A { // there will be another impl class
            @Override
            public String toString() { return "foobar"; }
        },
        ENUM_B;  // this might use the base enum impl as-is (or different impl, not 100% sure)
    }

```

either way, you get "one more level" in between Enum (interface type) and class that represents actual Enum values.

I hope this makes more sense.

ashleyfrieze:
@cowtowncoder something like this?:

```java
    private interface EnumIdentifier {
        @JsonValue
        String getId();
    }
    
    protected enum SubclassingEnum implements EnumIdentifier {
        ENUM_A {
            @Override
            public String getId() {
                return "A";
            }
        },
        ENUM_B {
            @Override
            public String getId() {
                return "B";
            }
        };
        
        @JsonCreator
        public static SubclassingEnum from(String id) {
            if ("A".equals("id")) {
                return ENUM_A;
            }
            if ("B".equals("id")) {
                return ENUM_B;
            }
            return null;
        }
    }
```

cowtowncoder:
@ashleyfrieze Yup.

ashleyfrieze:
@cowtowncoder  check out this version of the PR. I'll squash it before a final merge, but I've put this one in as a separate commit. Part of the challenge was making the ValueLocator recognise `enum` subclasses as being of type ID `SER_ENUM`.

What do you think?

ashleyfrieze:
@cowtowncoder I may have time for further revisions to this in about 15 hours - any feedback on this iteration?

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
```diff
@@ -75,4 +97,115 @@ public void writeValue(JSONWriter context, JsonGenerator g, Object value) throws
...
+            this._valueType = _valueType;
+            this._jsonValueMethod = _jsonValueMethod;
+        }
+
+        @Override
+        public void writeValue(JSONWriter context, JsonGenerator g, Object value) throws IOException {
+            try {
+                context.writeValue(_jsonValueMethod.invoke(value));
+            } catch (Exception e) {
+                throw new IOException("Cannot call JsonValue method", e);
```
cowtowncoder, line 118: I'd use Jackson jr specific exception here (`JSONObjectException`), not plain `IOException`.

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
```diff
@@ -75,4 +97,115 @@ public void writeValue(JSONWriter context, JsonGenerator g, Object value) throws
...
+        private static EnumJsonValueWriter of(Class<?> type) {
+            if (!isEnum(type)) {
+                return null;
+            }
+            return getJsonValueFunction(type, type);
+        }
+
+        private static EnumJsonValueWriter getJsonValueFunction(Class<?> type, Class<?> inspectionType) {
+            for (Method method : inspectionType.getDeclaredMethods()) {
+                if (method.getDeclaredAnnotation(JsonValue.class) != null &&
```
cowtowncoder, line 143: I would filter out non-static methods to avoid problems (or indicate an error for attempting).

Also not sure what handling should be for non-public methods; jackson-databind can force access but I don't remember if jackson-jr tries that or not.

cowtowncoder: Actually yes, `ValueReader/WriterLocator` do force access as needed. So should be done here too, probably.

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
```diff
@@ -75,4 +97,115 @@ public void writeValue(JSONWriter context, JsonGenerator g, Object value) throws
...
+                        method.getParameterCount() == 0 &&
+                        method.getReturnType().equals(String.class)) {
```
cowtowncoder, line 145: Also, note that:

    @JsonValue(false)

means "disabled", i.e. NOT to be used. Possibly not relevant with jackson-jr as the only use is via mix-ins -- but in theory should check "annotation.getValue()" being `true`.

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
```diff
@@ -75,4 +97,115 @@ public void writeValue(JSONWriter context, JsonGenerator g, Object value) throws
...
+                    return new EnumJsonValueWriter(type, method);
+                }
+            }
+
+            Class<?> superClass = inspectionType.getSuperclass();
+            if (superClass != null) {
```
cowtowncoder, line 151: These lead to some unnecessary work so might want to refactor so as not to check same supertypes multiple times.

jr-annotation-support/src/test/java/com/fasterxml/jackson/jr/annotationsupport/ASTestBase.java
```diff
@@ -11,8 +13,50 @@
...
+    /**
+     * enum that uses {@link JsonProperty} to alias values
+     */
     protected enum ABCRename { @JsonProperty("A1") A, @JsonProperty("B1") B, C; }
 
+    /**
+     * enum that uses a {@link JsonValue} method to alias its values and a {@link JsonCreator} to
+     * deserialize
+     */
+    protected enum ABCJsonValueJsonCreator
```
cowtowncoder, line 25: Probably does not need to be in test base class but actual test class? (unless it is shared across tests)

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/EnumReader.java
```diff
@@ -28,6 +28,21 @@ public EnumReader(Class<?> enumType,
         _byName = byName;
     }
 
+    /**
+     * Detect whether a class is an enum or a subclass of an enum - e.g. an anonymous inner class
+     * inside an enum
+     * @param type the type to inspect
+     * @return <code>true</code> if effectively an enum
+     */
+    public static boolean isEnum(Class<?> type) {
```
cowtowncoder, line 37: Why are these methods here, although used in different place? Probably belongs in caller, or a util-class of some type.

cowtowncoder:
@ashleyfrieze Looks good in general; added a few notes mostly regarding edge cases, placement of code.

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
ashleyfrieze: @cowtowncoder  - I can't quite see what you mean here. It should be pointed out that for enums, the hierarchy will be almost zero, so there's unlikely to be much work here. This is a recursive scan of the immediate class and its parents, with superclass, then interfaces, and it exits as early as possible. The only way I could imagine reducing work would be to add a `ConcurrentHashMap` to cache the discovered type information.

jr-annotation-support/src/test/java/com/fasterxml/jackson/jr/annotationsupport/ASTestBase.java
ashleyfrieze: With that in mind, I've moved `ABCRename` too and deleted the unused `ABC`

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/EnumReader.java
ashleyfrieze: Moved to a new utility class

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
ashleyfrieze: I have removed some unnecessary rechecking of `isEnum` on the `of` methods I created - they're only called by something that _knows_ the type is an enum

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
ashleyfrieze: Done - this might be an issue with illegal reflective access in later Java, though.

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
ashleyfrieze: I've written a test that demonstrates that one can add a `@JsonValue(false)` to a subclass override of an `@JsonValue(true)` method in order to disable that -  this means that you can turn off the `@JsonValue` behaviour, but it doesn't mean you can replace it with a different one.

jr-annotation-support/src/test/java/com/fasterxml/jackson/jr/annotationsupport/ASTestBase.java
```diff
@@ -2,17 +2,15 @@
...
 import com.fasterxml.jackson.jr.ob.JSON;
 
 import junit.framework.TestCase;
 
 public abstract class ASTestBase extends TestCase
 {
-    protected enum ABC { A, B, C; }
-
-    protected enum ABCRename { @JsonProperty("A1") A, @JsonProperty("B1") B, C; }
-
```
ashleyfrieze, line 15: Only used in the subclass. I originally added to this base class to follow the original pattern. Have tidied up a little.

ashleyfrieze:
Hey @cowtowncoder - review comments addressed. Please let me know what comes up when you go through this version.

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
cowtowncoder: Maybe I just misread recursion code here, mostly wrt interface detection. But I think I was confused about logic and there really isn't an issue there; although it is possible that an interface may be marked implemented at multiple levels that's not a big deal (even if gets scanned more than once).
So we are good.

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedValueRWModifier.java
cowtowncoder: That sounds good -- as  I suggested flexibility with this setting has mostly value with Jackson mix-ins and less so without. But thank you for covering the case there is, it is good to handle intended behavior.

cowtowncoder (approved):

cowtowncoder:
Oh. Just one more thing that I should have asked earlier: unless I have asked for it, I'd need CLA for this first contribution, from:

It only needs to be done once and is good for any and all future contributions.

The usual way is to fill the 1-page document, sign, scan/photo, email to `info` at fasterxml dot com.

Apologies for not mentioning this earlier.

ashleyfrieze:
@cowtowncoder CLA on its way to your inbox. Thanks for including my contribution.

PR #101, opened 2023-01-29
ci(deps): bump maven help plugin to 3.3.0

yeikel:


.github/workflows/main.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       run: ./mvnw -B -q -ff -ntp verify
     - name: Extract project Maven version
       id: projectVersion
-      run: echo "version=$(./mvnw org.apache.maven.plugins:maven-help-plugin:3.2.0:evaluate -DforceStdout -Dexpression=project.version -q)" >> $GITHUB_OUTPUT
+      run: echo "version=$(./mvnw org.apache.maven.plugins:maven-help-plugin:3.3.0:evaluate -DforceStdout -Dexpression=project.version -q)" >> $GITHUB_OUTPUT
```
pjfanning, line 49: you can remove the version number altogether - this appears to work

`./mvnw org.apache.maven.plugins:maven-help-plugin:evaluate -DforceStdout -Dexpression=project.version -q`

yeikel: That defaults to the latest version. The challenge is that it introduces [non-reproducible builds](https://reproducible-builds.org/)

yeikel: It is also not recommended by Maven itself. See https://github.com/apache/maven/blob/1968951a53cfefb75324f735579cbb68879ad5e0/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java#L1241-L1244

pjfanning: * this will not be accepted in the 2.14 branch - 2.15 is the dev branch
* this call is for pulling a version number out of the pom file and there is no documented bugs about this in the maven-help-plugin - this PR does not fix anything, does not improve the build in any way

yeikel: 1. When I created the pull request 2.14 was the latest branch and we can fix that easily. 2.15 was released around the same time
2. We can do that in a subsequent change. I disagree with you and believe that **considering the existing setup**, this PR does add value

cowtowncoder: Thank you @yeikel. Yes, I agree wrt using version number due to warning. And 2.15 is the recommended place, although for this particular change 2.14 would have been ok.

Issue #100, opened 2022-12-27
Add support for `java.time` (Java 8 date/time) types


sebastian-zero:
We found that jackson-jr is great in cases where small library size and fast initialization time is important. We use it in Android apps and AWS Lambdas. But one thing that is missing is support for date/time types as specified in JSR-310 specification.

I understand that this is currently only possible by implementing custom ValueReaders/ValueWriters. The documentation points to a blog post with some Groovy code that is not available anymore and can only be found through the [web archive](https://web.archive.org/web/20210805053523/https://www.andersaaberg.dk/2020/enable-support-for-java-time-with-jackson-jr/). 

Please consider adding support for Java 8 date/time types to jackson-jr, e.g. though an additional library/module that provides the required ValueReaders and ValueWriters.

cowtowncoder:
This sounds like a great idea @sebastian-zero! Also happy to hear this library works for one of the cases I intended it to (well, actually 2! :) ).

Unfortunately I probably won't have time to work on this directly (even if starting with the linked-to article which sounds like a great starting point!), but I hope that someone else could -- and if so I'll do my best to help them get additional Extension merged.
I prioritize work helping get PRs merged above almost everything else so if anyone is interested in tackling this, please do!

Shounaks:
Sounds like a easy contribution. How do I start?

cowtowncoder:
Initial implementation supporting just `LocalDateTime` now merged in for 2.17.

PR #105, opened 2023-07-09
Adding LocalDateTime Support via Extensions (Issue #100)

Shounaks:
relates to #100

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>com.fasterxml.jackson</groupId>
     <artifactId>jackson-base</artifactId>
-    <version>2.16.1</version>
+    <version>2.16.2-SNAPSHOT</version>
   </parent>
   <groupId>com.fasterxml.jackson.jr</groupId>
   <artifactId>jackson-jr-parent</artifactId>


jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/JSON.java
```diff
@@ -6,6 +6,7 @@
 
 import com.fasterxml.jackson.core.*;
 import com.fasterxml.jackson.core.io.SegmentedStringWriter;
+import com.fasterxml.jackson.core.json.PackageVersion;
```
pjfanning, line 9: I think you added this import by accident - can you remove it?

Shounaks: Done!

cowtowncoder:
This looks like a good starting point, but one quick note: instead of adding code to existing Maven sub-projects (general jr-objects, annotation-support), this would need to be a new sub-project/module.

Like `jr-extension-datetime` (for example). It'd depend on `jr-objects` and could be depended on (and registered) by projects that need it.
I'd prefer this over adding limited functionality in core; this allows plugging alternate implementations, and publishing new versions of datatype extensions quicker if necessary.

PR #111, opened 2024-02-11
Add extension `jackson-jr-extension-javatime` to support (some) Java 8 date/time types

Shounaks:
Sorry about previous tries, I believe this might suffice.
Regarding Issue #100


cowtowncoder:
Quick note: I fixed issue with CI failures (a regression in jackson-core), so now all tests pass as expected.

cowtowncoder:
First of all: thank you for contributing this, @Shounaks !
This sounds good overall; I'll need to go through it but I think the idea works well.
About the only immediate thing I noticed is that we need Moditect definition under

     ..../src/moditect/module-info.java

(see examples of f.ex jr-annotation-support/) 

Aside from that, one thing we need is CLA, from:

and it only needs to be sent once before the first contribution. Usually it's easiest to print, fill & sign, scan/photo, email to `cla` at fasterxml dot com.

Thank you again; I am looking forward to getting this merged for inclusion in 2.17.0 release!

pom.xml
```diff
@@ -26,6 +26,7 @@
     <module>jr-stree</module>
     <module>jr-annotation-support</module>
     <module>jr-all</module>
+    <module>jr-extension-datetime</module>
```
cowtowncoder, line 29: Let's move this above `jr-all` (since it needs to be built first: Maven can figure it out but just in case)

jr-extension-datetime/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+    <parent>
+        <groupId>com.fasterxml.jackson.jr</groupId>
+        <artifactId>jackson-jr-parent</artifactId>
+        <version>2.17.0-rc1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jr-extension-datetime</artifactId>
+
+    <properties>
+        <maven.compiler.source>20</maven.compiler.source>
```
cowtowncoder, line 15: Um, no, I don't think this makes sense. We should still have Java 8 base line.

cowtowncoder:
Ok, one more thing: I think main `README.md` should be modified to add brief explanation of the new extension. And there should be small `jr-extension-datetime/README.md` as well.

I will also remove JDK 20 references; package is still JDK 8 compatible for Jackson 2.x

jr-extension-datetime/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+    <artifactId>jackson-jr-extension-datetime</artifactId>
+    <packaging>bundle</packaging>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.fasterxml.jackson.jr</groupId>
+            <artifactId>jackson-jr-objects</artifactId>
+        </dependency>
+    </dependencies>
+
```
cowtowncoder, line 21: Ok, this needs bit more work: you can look at other `pom.xml`s like under `jr-retrofit2` or `jr-stree`. OSGi metadata is needed (for packaging as "bundle"), moditect generation.
And I think Gradle metadata.
I can help with these too.

README.md
```diff
@@ -40,6 +40,7 @@ Project is composed of multiple Maven sub-modules, each corresponding to a jar:
 * [jr-retrofit2](../../tree/master/jr-retrofit2) contains `jackson-jr` - based handlers for [Retrofit 2](https://square.github.io/retrofit/) library
     * Depends on `jackson-jr` and `Retrofit` API jars, and indirectly on `jackson-core`
 * [jr-annotation-support](../../tree/master/jr-annotation-support) contains extension with support for a subset of core [Jackson annotations](../../../jackson-annotations)
+* [jr-extension-datetime](../../tree/master/jr-extension-datetime) contains extension with support for `LocalDateTime`
```
cowtowncoder, line 43: Although it is just one type for now, could mention it is for supporting "a subset of Java 8 Date/Time types", here, and then in README.md for extension list `DateTime`.

jr-extension-datetime/README.md
```diff
@@ -0,0 +1,28 @@
+## Overview
+
+This module extends the functionalities of jackson-jr adding support for LocalDateTime.
```
cowtowncoder, line 3: And as per above, intent is to eventually support more Java 8 date/time types I think; not just this one.

cowtowncoder:
Looks good; if we can get CLA sorted out I think I should be able to merge this!

jr-extension-datetime/README.md
Shounaks: Okay, I have made the changes in readme. Now everything looks good 😄

Shounaks:
There is one more thing I would like to add, if you permit. that would be custom formatter support for this functionality
image
so using this, the developer would have a control on the format data is is coming on. I am sharing its patch file as well, feel free to tell me what you think about it.

[Adding_Custom_Formatter_Support_for_LocalDateTime.patch](https://github.com/FasterXML/jackson-jr/files/14290583/Adding_Custom_Formatter_Support_for_LocalDateTime.patch)

cowtowncoder:
Sure, the idea of configurability sounds reasonable. One thing I would suggest, however, is to not call formatter DEFAULT_FORMATTER and assume it is used for everything, since if and when adding new types they will need different formatters.

So I would probably add separate setter method for configuring `DateTimeFormatter` for `LocalDateTime`.

.... actually, I only now realize that the provider probably should be renamed too, not to be just for `LocalDateTime`. I think it should be something more generic.

jr-extension-datetime/src/main/java/com/fasterxml/jackson/jr/extension/datetime/JacksonLocalDateTimeExtension.java
```diff
@@ -0,0 +1,11 @@
+package com.fasterxml.jackson.jr.extension.datetime;
+
+import com.fasterxml.jackson.jr.ob.JacksonJrExtension;
+import com.fasterxml.jackson.jr.ob.api.ExtensionContext;
+
+public class JacksonLocalDateTimeExtension extends JacksonJrExtension {
```
cowtowncoder, line 6: As per my other note, should probably rename these NOT to refer to `LocalDateTime` since we probably do not want 12 different extensions for Java 8 added date/time types.

So something like `JacksonJavaTimeExtension`? And similarly for providers etc.

Whether extension should be `jr-extension-datetime` or `jr-extension-javatime` (latter just because JDK package is `java.time`) I leave up to you.

Shounaks: yes that makes much more sense, I have changed the name of the module to `jr-extension-javatime`, and implemented the setter method required to set custom formatters:
image
so now , if anyone wants to specify any custom one, it will just override the existing one, at the same time, providing with a 'sane default' choice, as before.

cowtowncoder (approved):

cowtowncoder:
First of all: thank you @Shounaks for contributing this! It will be in Jackson 2.17.0 rc1 to be released soon.

One thing: I Merged before remembering to ask or one thing -- basic set of unit tests would be nice, if possible (via different PR)?

PR #117, opened 2024-02-17
Fix #116: add read/write support for `java.nio.file.Path`

### LINKED ISSUE #116: Add read/write support for `java.nio.file.Path`
(follow-up to #112)

Looks like it'd be useful to support JDK 1.7+ `java.nio.file.Path` value type.
Let's add very basic reader, writer support.

cowtowncoder:


jhonnen:
Should a `Path` really be serialised as an `URI`? I would have assumed you could interchange `Path` and `File` while reading and writing the same JSON string.

cowtowncoder:
@jhonnen I was looking at what Jackson databind does and followed that. But I am open to counter arguments -- and PRs would be welcome for changed behavior (and tests).
We can easily change it now before this gets released in 2.17.

jhonnen:
Following Jackson databind is a good point. I just think it's confusing that
```
{ "x": "/some/path" }
```
can only be deserialized as `File` but not as `Path`.

cowtowncoder:
@jhonnen As I said, I am totally open to alternative handling, since I do not actually have use for this myself.
So if you feel different handling makes sense, I would really appreciate a PR for improvement (just so I do not misunderstand some point; unit test showing exactly expected behavior).

And in fact use of `Path.toString()` would sort of go with least-surprise approach (even if less stable for round-tripping).

Issue #93, opened 2022-01-20
Skip serialization of `groovy.lang.MetaClass` values to avoid `StackOverflowError` 


chashnikov:
Try serializing empty Groovy class using Groovy 3.0.9 and jackson-jr-objects:2.13.1:
```groovy
import com.fasterxml.jackson.jr.ob.JSON
import groovy.transform.CompileStatic

class MyBean {}

@CompileStatic
class JsonTest {
    static void main(String[] args) {
        println JSON.std.asString(new MyBean())
    }
}
```
It'll fail with `StackOverflowError`:
```stacktrace
	at com.fasterxml.jackson.jr.ob.impl.JSONWriter.writeBeanValue(JSONWriter.java)
	at com.fasterxml.jackson.jr.ob.impl.BeanWriter.writeValue(BeanWriter.java)
	at com.fasterxml.jackson.jr.ob.impl.JSONWriter._writeValue(JSONWriter.java)
	at com.fasterxml.jackson.jr.ob.impl.JSONWriter.writeBeanValue(JSONWriter.java)
	at com.fasterxml.jackson.jr.ob.impl.BeanWriter.writeValue(BeanWriter.java)
	at com.fasterxml.jackson.jr.ob.impl.JSONWriter._writeValue(JSONWriter.java)
	at com.fasterxml.jackson.jr.ob.impl.JSONWriter.writeBeanValue(JSONWriter.java)
	at com.fasterxml.jackson.jr.ob.impl.BeanWriter.writeValue(BeanWriter.java)
```
This happens because every class file generated from Groovy class contains `getMetaClass` method, which contains data with meta-information with cyclic references inside. The method is marked with `@java.beans.Transient` annotation, so if serialization automatically skips such properties, the problem will be solved.

cowtowncoder:
I think it is possible for users to implement something like this, similar to `jr-annotation-support`.

Now, this could make useful addition to `jr-annotation-support` itself, but there is one practical problem: annotation was added in JDK 7, and `jackson-jr` only requires JDK 6 as of Jackson 2.13. We could consider increasing baseline to JDK 8 in which case it'd be possible to add support.

If anyone wanted to provide a PR for addition here, I'd be happy to help to see if JDK baseline can be improved.

chashnikov:
> annotation was added in JDK 7, and jackson-jr only requires JDK 6 as of Jackson 2.13

[...]

PR #118, opened 2024-02-17
Skipping over groovy metadata class + groovy test.

Shounaks:
Issue: #93 
Description: As per our discussion, we will be skipping serialization of Groovy related classes , i have added a sample groovy test case as well!

Reference taken from: https://github.com/FasterXML/jackson-databind/blob/db95863d6f48cb5695bf8e14af1e7af2ece8e52e/src/main/java/com/fasterxml/jackson/databind/util/BeanUtil.java#L205


jr-objects/src/test/java/com/fasterxml/jackson/jr/GroovyTest.groovy
```diff
@@ -0,0 +1,21 @@
+package com.fasterxml.jackson.jr
+
+import com.fasterxml.jackson.jr.ob.JSON
+import com.fasterxml.jackson.jr.ob.TestBase
+
+class GroovyTest extends TestBase {
+
+    void testSimpleObject() throws Exception {
+        var data = JSON.std.asString(new MyClass())
+        var expected = "{\"aDouble\":0.0,\"aStr\":\"stringData\",\"anInt\":0,\"metaClass\":{}}";
```
cowtowncoder, line 10: I think ideally we would not serialize this field at all, to avoid problems on deserialization.

cowtowncoder:
Ok, so, this would indeed avoid cyclic dependency but I have 2 concerns.

One is that it would prevent any all types in `groovy.lang` (and sub-packages) from being (de)serialized, not just metadata. Not sure this is a big problem but could be problematic.

Second: jackson-databind prevents serialization of all instances accessed via `getMetaClass()` that return type under `groovy.lang`, which seems more targeted and prevents inclusion of any value.
Approach here will still leave empty Object. To me, removal would seem preferable; and one that specifically targets property that is the problem.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanPropertyIntrospector.java
```diff
@@ -1,189 +1,149 @@
 package com.fasterxml.jackson.jr.ob.impl;
 
+import com.fasterxml.jackson.jr.ob.impl.POJODefinition.Prop;
+import com.fasterxml.jackson.jr.ob.impl.POJODefinition.PropBuilder;
+
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
-import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.util.Arrays;
```
cowtowncoder, line 9: As per Jackson code style import order for non-statics should be:

1. JDK imports
2. 3rd party imports, if any
3. Core jackson imports if any
4. Imports from package itself

Shounaks:
Okay Will create keep this PR for just precise changes, and then create a seperate one for Code Cleanup. if its fine.

Shounaks:
Done! 👍👍

jr-objects/src/test/java/com/fasterxml/jackson/jr/GroovyTest.groovy
Shounaks: Now the Meta class will not serialize

jr-objects/src/test/java/com/fasterxml/jackson/jr/GroovyTest.groovy
```diff
@@ -0,0 +1,58 @@
+package com.fasterxml.jackson.jr
+
+import com.fasterxml.jackson.jr.ob.JSON
+import com.fasterxml.jackson.jr.ob.TestBase
+
+class GroovyTest extends TestBase {
+
+    void testSimpleObject() throws Exception {
+        var data = JSON.std.asString(new MyClass())
+        var expected = """{"AAAAA_A_Field_Starting_With_Two_Capital_Letters":"XYZ","aDouble":0.0,"aPublicInitializedInteger":56,"aPublicInitializedIntegerObject":1516,"aPublicUninitializedInteger":0,"anInitializedIntegerObject":1112,"anInitializedPublicString":"stringData","anInitializedString":"ABC","anInteger":0,"anIntegerWithValue":12}"""
```
cowtowncoder, line 10: Was about to ask about `var` and triple quotes wrt JDK 8 before realizing this is Groovy :)

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanPropertyIntrospector.java
```diff
@@ -182,4 +177,13 @@ private static String decap(String name) {
...
     }
 
+    /**
+     * Another helper method to deal with Groovy's problematic metadata accessors
+     *
+     * @implNote Groovy MetaClass have cyclic reference, and hence the class containing it should not be serialised without
+     * either removing that reference, or skipping over such references.
+     */
+    protected static boolean isGroovyMetaClass(Class<?> clazz) {
+        return clazz.getName().startsWith("groovy.lang");
```
cowtowncoder, line 187: This needs to check for class name, not just package, right?

Shounaks: Initially i implemented it to check for just MetaClassImpl, but after checking jackson-databind, i just replicated its exclusion.

https://github.com/FasterXML/jackson-databind/blob/db95863d6f48cb5695bf8e14af1e7af2ece8e52e/src/main/java/com/fasterxml/jackson/databind/util/BeanUtil.java#L205

is this correct? or should we go for targeted exclusions?

jr-objects/src/test/java/com/fasterxml/jackson/jr/GroovyTest.groovy
cowtowncoder: One question tho: how does this test get run? It doesn't look like it runs with `./mvnw clean test` from main level...

cowtowncoder: As per 

https://ppettitau.wordpress.com/2013/04/04/running-groovy-junit-tests-with-surefire-in-a-javamaven-project/

that (moving test) is one of 3 things needed, fwtw.

cowtowncoder:
Ok: I was able to make groovy test actually run -- but it does not appear to trigger cyclic dependency even without change to `BeanPropertyIntrospector`. :-(

So not 100% sure how to trigger the problem being fixed, still.

Shounaks:
> Ok: I was able to make groovy test actually run -- but it does not appear to trigger cyclic dependency even without change to `BeanPropertyIntrospector`. :-(
> 
> So not 100% sure how to trigger the problem being fixed, still.

image
I am able to replicate failure if i comment out these 2 lines of code.
image

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanPropertyIntrospector.java
Shounaks: understood, I missed this code, will implement in a similar fashion.

cowtowncoder:
@Shounaks Does not fail for me if I change `isGroovyMetaClass()` to return `false`: everything passes without problem. Test class is executed (I can verify with `Assert.fail()`).

Could it be you are commenting out more than the new check, so that pre-existing checks for missing type or `Object` are not done?

cowtowncoder:
Oh wait. I _can_ actually reproduce the issue via IDE (Eclipse) after installing Groovy plugin.

It just does not fail with Maven from command-line. This is bizarre...

cowtowncoder:
Ok. I really wish Maven was able to reproduce (to guard against regression), but it is what it is. Will merge in for 2.17.0(-rc1).

Shounaks:
> Oh wait. I _can_ actually reproduce the issue via IDE (Eclipse) after installing Groovy plugin.
> 
> It just does not fail with Maven from command-line. This is bizarre...

yeah, this is strange. Never would've guessed this can be a point of failure. 😆

cowtowncoder:
Yeah more commonly it's the opposite: some problem with IDE, works as expected via Maven from shell.
I wonder if part of this might be compiler plug-in settings: for jackson-jr we are compiling with minimum debug information to reduce jar size. Not sure that is related, but I did notice that compiling from shell vs IDE compiling has some effect. Could of course just be wrt Groovy integration too.

PR #119, opened 2024-02-20
Change post #116 to potentially change serialization of `java.nio.file.Path`

cowtowncoder:


cowtowncoder:
@jhonnen So, this would change behavior to, I think, what you are asking. WDYT?
One more possibility would be addition of a new `JSON.Feature` to control behavior (choose one of 2, I suspect).

Either way, I am hoping to finalize handling soon, for upcoming 2.17.0-rc1 release.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
```diff
@@ -278,7 +278,8 @@ protected Path _readPath(JsonParser p) throws IOException {
         }
         String v = p.getValueAsString();
         try {
-            return Paths.get(new URI(v));
+//            return Paths.get(new URI(v));
+            return Paths.get(new File(v).toURI());
```
jhonnen, line 282: why not simplify to
```suggestion
            return Paths.get(v);
```

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/JSONWriter.java
```diff
@@ -789,7 +789,8 @@ protected String dateToString(Date v) {
      * @since 2.17
      */
     protected String pathToString(Path value) {
-        return value.toUri().toString();
+//        return value.toUri().toString();
+        return value.toString();
```
jhonnen, line 793:
```suggestion
        return value.toString();
```

jhonnen (approved):
that's the behavior I expected 👍

Issue #25, opened 2015-06-02
Add support for single-int Constructors


cowtowncoder:
Currently single-String and single-long/Long constructors are accepted. Perhaps we should consider allowing single-int/Integer ones too.

Shounaks:
This issue can be now be closed.

PR #121, opened 2024-02-22
Adding Single Int Constructor Support (Issue #25)

Shounaks:
(Issue #25)
this already had a failing test (we should add this tag)


jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
```diff
@@ -34,19 +34,22 @@ public class BeanReader
...
-            Constructor<?> defaultCtor, Constructor<?> stringCtor, Constructor<?> longCtor,
+            Constructor<?> defaultCtor, Constructor<?> stringCtor, Constructor<?> longCtor, Constructor<?> intCtor,
             Set<String> ignorableNames, Map<String, String> aliasMapping)
     {
         super(type);
         _propsByName = props;
         _defaultCtor = defaultCtor;
         _stringCtor = stringCtor;
         _longCtor = longCtor;
+        _intCtor = intCtor;
```
Shounaks, line 51: I believe propogation of HashMap<FIRST_ARG,CONSTRUCTOR> should be a better alternative, to support future items. instead of adding individually in all the following files.

jr-objects/src/test/java/com/fasterxml/jackson/jr/failing/ReadWithCtors25Test.java
```diff
@@ -6,12 +6,12 @@
 // for [jackson-jr#25], allowing single-int constructors
 public class ReadWithCtors25Test extends TestBase
```
Shounaks, line 7: need to move this to passing tests, once the implementation is approved

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
cowtowncoder: I agree with the general idea (could use `EnumMap` too), but how about going even more OO and have `Constructors` (or whatever its named) container that encapsulates these details?

cowtowncoder (commented):
Ok, yes, this makes sense -- but I think the idea of creating separate container class for different constructors makes sense: can then pass that object instead of 4 constructors (plus obv easier extension in future, as necessary).

With that I'll be happy to merge this!

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
Shounaks: I tried using enum map, but that would require us explicitly manage a list of supported items, with the code getters and setters; a problem not present in HashMap.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ConstructorDefinition.java
```diff
@@ -0,0 +1,29 @@
+package com.fasterxml.jackson.jr.ob.impl;
```
cowtowncoder, line 1: Um, sorry, but this looks unnecessary complicated: instead of `Map`, suppliers etc, let's just stick to plain old typed fields. Map adds overhead while not improving readability.

cowtowncoder:
Thank you for contributing this PR! I created slightly modified one based on my preferences wrt Constructor collector, merged: since it is based on your code I put you as the author in release notes.

I'll close this PR as done.

Issue #94, opened 2022-02-22
Support for serializing Java Records


jam01:
Hey all, I did not find any explicit claim that Java Records were supported so I decided to give it a shot. It seems to me that they are not, given the following snippet:

```
    @Test
    public void testRecord() throws IOException {
        var json = """
                {
                    "string":"foo",
                    "object": {
                        "foo": "bar"
                    }
                }
                """;

        var rec = JSON.std.beanFrom(Cow.class, json);
        System.out.println(rec);
    }

    public record Cow(String message, Map<String, String> object) {}
```

results in:

```
com.fasterxml.jackson.jr.ob.JSONObjectException: Failed to create an instance of com.example.Cow due to (java.lang.IllegalStateException): Class com.example.Cow does not have default constructor to use
 at [Source: (String)"{
    "message":"foo",
    "mappy": {
        "foo": "bar"
    }
}
"; line: 1, column: 1]

	at com.fasterxml.jackson.jr.ob.JSONObjectException.from(JSONObjectException.java)
	at com.fasterxml.jackson.jr.ob.impl.BeanReader.read(BeanReader.java)
	at com.fasterxml.jackson.jr.ob.impl.JSONReader.readBean(JSONReader.java)
	at com.fasterxml.jackson.jr.ob.JSON.beanFrom(JSON.java)
	at com.example.AppTest.testRecord(AppTest.java:30)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:149)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:140)
[...]

PR #122, opened 2024-02-22
Adding Supports for Records (Groovy) + Passing Record Test.

Shounaks:
Issue : support for records #94

This will enable the introspector to catch Groovy generated getters, and serialize the records.


jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanPropertyIntrospector.java
```diff
@@ -142,6 +143,13 @@ private static void _introspect(Class<?> currType, Map<String, PropBuilder> prop
                         name = decap(name.substring(2));
                         _propFrom(props, name).withIsGetter(m);
                     }
+                } else {
+                    // This will allow getters with field name as their getters, like the ones generated by Groovy
+                    boolean isDirectNameGetter = Arrays.stream(currType.getDeclaredFields()).map(Field::getName).anyMatch(f -> f.equals(m.getName()));
```
Shounaks, line 148: i believe if we get a list of fields & check if there is a field with the same name as current method, then we need to add it as a prop

, this can be optimized, but as of now, its working.

cowtowncoder: I think this capability needs to be behind `Json.Feature` flag since while it makes sense wrt Records (and probably other value types), it can also be unexpected for other users. That is, it changes compatibility.

Other than that it'd make sense to build a map from name to `Field` in loop above, to avoid linear scanning for every potential "non-get-getter".

cowtowncoder:
Quick note: if this is meant for Groovy records specifically, should probably file a new separate issue -- #94 is, I think, for Java 17 Record type and require immutability including construction by multi-argument constructor.
This seems to be just for automatically recognizing "non-get getters" (so, `name()` for property `"name"` if (and only if) there is matching Field), which would also work for Java Records, but wouldn't be full Record support in my mind.

I also think this should be behind a new `JSON.Feature` (DETECT_FIELD_MATCHING_GETTERS? or some better name, ideally :) ) just because it is behavioral change that might not be ideal for some users.
(I think I'd be ok but it defaulting to `true`, just as long as it is possible to disable it for compatibility reasons).

jr-objects/src/test/groovy/com/fasterxml/jackson/jr/passing/GroovyTest.groovy
```diff
@@ -0,0 +1,68 @@
+package com.fasterxml.jackson.jr.passing
```
cowtowncoder, line 1: Is the new `passing` needed to avoid split packages or... ?

If necessary, maybe use `com.fasterxml.jackson.jr.ob.groovy`.

Shounaks: my mistake, removed in latest commits.

Shounaks:
Yes, this will fix Java14's Records as well as Groovy Records.

> This seems to be just for automatically recognizing "non-get getters" (so, name() for property "name" if (and only if) there is matching Field), which would also work for Java Records, but wouldn't be full Record support in my mind.

Okay so, how should we go on to check which object is a record and which is not, without having access to native internal methods?

I tried to do some digging, and found out this:
image
but we dont really support JDK 16..
image
and i couldnt get to the end of how this native method is implemented. (no experience in this field sadly..)

One immediate way that i can think of is to check constructor parameters matching the field names .
image
but i am not sure how it will interact with current code. Might need your help here @cowtowncoder .

cowtowncoder:
It is possible to detect "Record-ness" of a type, even from JDK 8, but that does get tricky: jackson-databind does that. Actually detecting something is Record is relatively easy: `src/main/java/com/fasterxml/jackson/databind/util/ClassUtil.java`: has this:

```java
    public static boolean isRecordType(Class<?> cls) {
        Class<?> parent = cls.getSuperclass();
        return (parent != null) && "java.lang.Record".equals(parent.getName());
    }
```

which we could re-purpose. Same might work for Groovy types.

Anything fancier (finding canonical constructor, declared properties) gets trickier, but just knowing something is a Record type is quite doable.

cowtowncoder:
I think that we have a choice, so either:

1. Use a `JSON.Feature` to enable detection, regardless of type (whether it's Record or not)
2. Auto-detect Records and apply different rules

I guess one could combine things too, to basically default to auto-detecting via (2), but allowing disabling by (1). Not sure how necessary that is.

cowtowncoder:
Ok, couple of additional notes:

1. Groovy test seems to fail for some reason in CI
2. Not sure how to test Java Records: jackson-databind uses different profiles to optionally included Java 17+ tests, that is a possibility -- but baseline for Jackson 2.x jackson-jr is JDK 8 so they can't be added in main `src/test/java`
3. ... although, since jackson-jr is multi-module Maven project, it potentially could have separate test-only package that runs after building `jr-objects`. Either way, gets bit complicated

We could, however, also test case of "not Record but has field-name matching getter methods" as surrogates for Records. This assuming we do not want to limit this feature to just Java/Groovy Record types.

Shounaks:
> I think that we have a choice, so either:
1. Use a JSON.Feature to enable detection, regardless of type (whether it's Record or not)
2. Auto-detect Records and apply different rules
I guess one could combine things too, to basically default to auto-detecting via (2), but allowing disabling by (1). Not sure how necessary that is.

I believe 1st choice is better, since we have the knob to turn this feature off, when working with other libraries,

Shounaks:
This failing test, it seems, that the since, the getter is public, we are getting the values, and hence we are able to serialize these items. looks like correct behavior. can you confirm once again? so that I will remove this test. or this is the expected behavior?

I didnt really think about `JsonAutoDetect.Visibility.PROTECTED_AND_PUBLIC` when implementation of this feature. how should these features interact with that?

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanPropertyIntrospector.java
```diff
@@ -145,6 +149,15 @@ private static void _introspect(Class<?> currType, Map<String, PropBuilder> prop
                         name = decap(name.substring(2));
                         _propFrom(props, name).withIsGetter(m);
                     }
+                } else if(USE_FIELD_NAME_GETTERS.enabledByDefault()){
```
cowtowncoder, line 152: As with other comment, `enabledByDefault()` is not to be used here -- we need actual configuration settings.

cowtowncoder:
Quick request: would it be possible to extract change to add new test module for Groovy classes into separate PR? We already added `jr-objects/src/test/groovy/com/fasterxml/jackson/jr/GroovyTest.groovy` and that could now be moved to the new module; and that PR could be merged before figuring out details of actual feature.

Shounaks:
sounds good, let me do that.

Shounaks:
> Quick request: would it be possible to extract change to add new test module for Groovy classes into separate PR? We already added `jr-objects/src/test/groovy/com/fasterxml/jackson/jr/GroovyTest.groovy` and that could now be moved to the new module; and that PR could be merged before figuring out details of actual feature.

Created a new PR: #128

Shounaks:
Closing this PR, will start with clean slate, with minimal changes, and formatting.

PR #124, opened 2024-02-25
Fix #48: add `java.util.Date/Calendar` Extension Support

### LINKED ISSUE #48: JSONObjectException when reading calendar field
Hi,

I'm writing a list of instances of my class that has a calendar field.

```
File file = new File(context.getFilesDir(), fileName);
try {
    JSON.std.write(listOfObjects, file);
```
Writing works just fine, but when reading the created json, I receive the following exception:

`com.fasterxml.jackson.jr.ob.JSONObjectException: Can not get long numeric value from JSON (to construct java.util.Calendar) from JSON String
`

I can fix this easily by adding the `JSON.Feature.WRITE_DATES_AS_TIMESTAMP`to the writer, but I find it rather odd that jackson-jr is not able to read what it has written with complete default configuration.

I'm using `'com.fasterxml.jackson.jr:jackson-jr-objects:2.7.4'`

ktxhkpnwbd:
Issue: #48 
Providing Support for java.util.Date, via Extensions.

### DIFF
--- a/jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/CalendarValueReader.java
+++ b/jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/CalendarValueReader.java
@@ -1,0 +1,60 @@
+package com.fasterxml.jackson.jr.extension.javatime;
+
+import java.io.IOException;
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.util.Calendar;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.jr.ob.JSONObjectException;
+import com.fasterxml.jackson.jr.ob.api.ValueReader;
+import com.fasterxml.jackson.jr.ob.impl.JSONReader;
+
+/**
+ * {@link ValueReader} for {@link java.util.Calendar}. Like {@link DateValueReader}
+ * reading is lenient: it accepts both numeric timestamps (milliseconds since the
+ * Unix epoch) and textual values parsed using the configured {@link DateFormat}.
+ *
+ * @since 2.23
+ */
+public class CalendarValueReader extends ValueReader {
+    private final DateFormat _dateFormat;
+
+    public CalendarValueReader(DateFormat dateFormat) {
+        super(Calendar.class);
+        _dateFormat = dateFormat;
+    }
+
+    @Override
+    public Object read(JSONReader reader, JsonParser p) throws IOException {
+        if (p.hasToken(JsonToken.VALUE_NULL)) {
+            return null;
+        }
+        // Always accept numeric timestamps, regardless of configured format
+        if (p.hasToken(JsonToken.VALUE_NUMBER_INT)) {
+            return _fromMillis(p.getLongValue());
+        }
+        final String text = p.getValueAsString();
+        if (text == null || text.isEmpty()) {
+            return null;
+        }
+        try {
+            final long millis;
+            // DateFormat is not thread-safe
+            synchronized (_dateFormat) {
+                millis = _dateFormat.parse(text).getTime();
+            }
+            return _fromMillis(millis);
+        } catch (ParseException e) {
+            throw new JSONObjectException("Failed to parse `java.util.Calendar` value '"
+                    + text + "': " + e.getMessage(), e);
+        }
+    }
+
+    private static Calendar _fromMillis(long millis) {
+        Calendar c = Calendar.getInstance();
+        c.setTimeInMillis(millis);
+        return c;
+    }
+}
--- a/jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/DateValueReader.java
+++ b/jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/DateValueReader.java
@@ -1,0 +1,53 @@
+package com.fasterxml.jackson.jr.extension.javatime;
+
+import java.io.IOException;
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.util.Date;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.jr.ob.JSONObjectException;
+import com.fasterxml.jackson.jr.ob.api.ValueReader;
+import com.fasterxml.jackson.jr.ob.impl.JSONReader;
+
+/**
+ * {@link ValueReader} for {@link java.util.Date}. Reading is lenient: it accepts
+ * both numeric timestamps (milliseconds since the Unix epoch) and textual values
+ * parsed using the configured {@link DateFormat}.
+ *
+ * @since 2.23
+ */
+public class DateValueReader extends ValueReader {
+    private final DateFormat _dateFormat;
+
+    public DateValueReader(DateFormat dateFormat) {
+        super(Date.class);
+        _dateFormat = dateFormat;
+    }
+
+    @Override
+    public Object read(JSONReader reader, JsonParser p) throws IOException {
+        if (p.hasToken(JsonToken.VALUE_NULL)) {
+            return null;
+        }
+        // Always accept numeric timestamps, regardless of configured format
+        if (p.hasToken(JsonToken.VALUE_NUMBER_INT)) {
+            return new Date(p.getLongValue());
+        }
+        final String text = p.getValueAsString();
+        if (text == null || text.isEmpty()) {
+            return null;
+        }
+        try {
+            // DateFormat is not thread-safe; readers get a private clone but we
+            // still guard against reuse within a single reader instance.
+            synchronized (_dateFormat) {
+                return _dateFormat.parse(text);
+            }
+        } catch (ParseException e) {
+            throw new JSONObjectException("Failed to parse `java.util.Date` value '"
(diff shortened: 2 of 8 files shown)


jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/date/DateValueReader.java
```diff
@@ -0,0 +1,25 @@
...
+import com.fasterxml.jackson.jr.ob.impl.JSONReader;
+
+import java.io.IOException;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public class DateValueReader extends ValueReader {
+    protected DateValueReader() {
+        super(Date.class);
```
cowtowncoder, line 14: Technically this is not JAva 8 date/time type... so probably should be in core `jr-objects`?

But also, it needs to support deserialization from Integer numbers (timestamp).

cowtowncoder:
Ok, I think this could use refactoring:

1. Not sure if `java.util.Date` handling should be added here: core `jr-objects` supports it with timestamp; should not simply overwrite it without configurability
2. No need to add new extension, just extend `JacksonJrJavaTimeExtension`
3. No need to add new `ReaderWriterProvider`, just extend `JavaTimeReaderWriterProvider`

I can do some of this.

cowtowncoder:
Oh shoot. I should have read it more carefully: this ONLY adds textual read/write for `java.util.Date`. I was hoping more support for Java 8 date/time types.

I think addition of more features would be ok but should be part of same extension (not separate one) and require configuration (by default leaving serialization as timestamp) -- it's ok to accept textual value by default tho.
However, there is then the question of how to specify format used, and how to (re)configure.

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/JavaTimeReaderWriterProvider.java
```diff
@@ -20,11 +25,26 @@
...
+
     private ZoneId _localZoneId;
-    
+
+    /**
+     * Optional format for {@link java.util.Date} serialization. When left
+     * {@code null}, dates are written as numeric timestamps (milliseconds since
+     * the Unix epoch); reading always accepts both numeric timestamps and
+     * textual values.
+     */
```
cowtowncoder, line 48: `@since 2.23`

cowtowncoder:
Should/could this also support `java.util.Calendar`, similarly?

cowtowncoder (commented):
(edited from Claude review)

### Correctness — looks good

- The core `ValueWriterLocator` change is the right fix and is genuinely necessary: without it a provider-supplied `ValueWriter` is only consulted at the top level, never for **bean properties**. The new branch now mirrors `_findPOJOSerializationType` and the reader side (`ValueReaderLocator._createReader`). The tests exercise exactly this path (Date/Calendar as a bean field), so it's load-bearing.
- `DateValueReader`/`CalendarValueReader` synchronize on `_dateFormat`. Correct and necessary — the reader/writer instances are cached and shared across threads, each holding a single (non-thread-safe) `DateFormat`. Cloning at resolution time + `synchronized` on use is the right combination.
    - HOWEVER! Synchronized here is REALLY bad for concurrent performance (see comment inline for more)

### Medium — need to resolve before merge

**1. The extension silently overrides the core `WRITE_DATES_AS_TIMESTAMP` feature.** Core already handles `Date`/`Calendar` (`SER_DATE`/`SER_CALENDAR` in `JSONWriter`, gated on `WRITE_DATES_AS_TIMESTAMP`, which defaults to `false` → ISO text). Once the javatime extension is registered, its writer takes over completely and that core Feature becomes a no-op for Date/Calendar. Two consequences:
- Registering the extension *flips the default* for `Date` from ISO-8601 text (core default) to numeric timestamp (extension default) — a surprising behavior change just from adding the extension.
- `JSON.builder().with(Feature.WRITE_DATES_AS_TIMESTAMP)` (or `.without(...)`) plus the extension has no effect.

This is the "should not simply overwrite it without configurability" concern from the original review note. The PR adds `withDateFormat(...)`, but doesn't reconcile with the existing flag.
Needs to honor `WRITE_DATES_AS_TIMESTAMP` in the new writers.

### Low — minor

**2.** `Date.class.isAssignableFrom(type)` also captures `java.sql.Date`/`java.sql.Timestamp`. `DateValueReader` always returns a plain `java.util.Date`, so a bean property typed `java.sql.Date` would get a `ClassCastException` on set. The writer is harmless (only reads `getTime()`), but the reader is a real (if niche) regression vs. not intercepting those subtypes. Worth a guard or a note.

**3.** `CalendarValueReader._fromMillis` uses `Calendar.getInstance()` — system default locale/timezone, so round-tripping loses the original zone. Fine for millis-equality (all the test asserts), and matches typical databind behavior, but worth a comment.

**4.** Silent nulls on odd input: for a non-null, non-numeric token that yields a null/empty `getValueAsString()` (e.g. `START_OBJECT`), the readers quietly return `null` rather than erroring.

**5.** Strictness mismatch with the rest of the extension: the Java 8 types use a deliberately "forgiving" `DateTimeFormatter`, while Date/Calendar default reads use a strict `SimpleDateFormat("…SSSXXX")` — input like `2024-02-25T01:52:00Z` (no millis) fails to parse.

**6.** Comment nit in `DateValueReader`: "readers get a private clone but we still guard against reuse within a single reader instance" — the `synchronized` actually guards cross-thread sharing of the one cached reader instance, not reuse within a single instance.

### Tests
Good coverage: timestamp round-trip, textual round-trip, lenient cross-reads, and null handling, for both Date and Calendar. The one gap (tied to #1) is no test asserting interaction with `WRITE_DATES_AS_TIMESTAMP`.

The main item to settle is **#1**.

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/CalendarValueReader.java
```diff
@@ -0,0 +1,60 @@
...
+    public Object read(JSONReader reader, JsonParser p) throws IOException {
+        if (p.hasToken(JsonToken.VALUE_NULL)) {
+            return null;
+        }
+        // Always accept numeric timestamps, regardless of configured format
+        if (p.hasToken(JsonToken.VALUE_NUMBER_INT)) {
+            return _fromMillis(p.getLongValue());
+        }
+        final String text = p.getValueAsString();
+        if (text == null || text.isEmpty()) {
```
cowtowncoder, line 39: Is this consistent with handling elsewhere? Should we quietly coerce empty String to `null`?

It seems better to fail with:

1. `null` case for wrong (non-String, non-Number) JSON token
2. For empty String just try parsing, fail.

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/CalendarValueReader.java
```diff
@@ -0,0 +1,60 @@
...
+            return null;
+        }
+        try {
+            final long millis;
+            // DateFormat is not thread-safe
+            synchronized (_dateFormat) {
```
cowtowncoder, line 45: This is really bad for concurrent access... Single-threaded for ALL parsing?
Not sure how to solve that -- maybe need to `clone()` instance before use; not optimal but better for concurrency.

Or alternatively try to store per-thread instance on `JSONReader` (there is one `JSONReader` per `Json.beanFrom()` call ). But that may get complicated too.

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/CalendarValueReader.java
cowtowncoder: For (1) case (not `JsonToken.VALUE_STRING`) could just use

```
            default:
                throw new JSONObjectException("Failed to bind `java.util.Calendar` from value: "+
                        _tokenDesc(p));
```

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/DateValueReader.java
```diff
@@ -0,0 +1,53 @@
...
+import com.fasterxml.jackson.jr.ob.impl.JSONReader;
+
+/**
+ * {@link ValueReader} for {@link java.util.Date}. Reading is lenient: it accepts
+ * both numeric timestamps (milliseconds since the Unix epoch) and textual values
+ * parsed using the configured {@link DateFormat}.
+ *
+ * @since 2.23
+ */
+public class DateValueReader extends ValueReader {
```
cowtowncoder, line 21: naming: `JavaUtilDateValueReader`

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/DateValueWriter.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * {@link ValueWriter} for {@link java.util.Date}. By default (when no
+ * {@link DateFormat} is configured) values are written as numeric timestamps
+ * (milliseconds since the Unix epoch); when a format is supplied they are
+ * written as textual values.
+ *
+ * @since 2.23
+ */
+public class DateValueWriter implements ValueWriter {
```
cowtowncoder, line 19: naming: `JavaUtilDateValueWriter`

PR #125, opened 2024-02-25
Using Stream Builder to deserialization of int[]

Shounaks:
Issue: #7 
we can use IntStream.Builder and iteratively add tokens. I don't know if its wasteful or not... 😥

### DIFF
--- a/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
+++ b/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
@@ -294,12 +294,12 @@ public class SimpleValueReader extends ValueReader
     }
 
     protected int[] _readIntArray(JsonParser p) throws IOException {
-        if (p.getCurrentToken().equals(JsonToken.START_ARRAY)) {
+        if (JsonToken.START_ARRAY.equals(p.currentToken())) {
             p.nextToken();
         }
 
         final IntStream.Builder builder = IntStream.builder();
-        while (!p.getCurrentToken().equals(JsonToken.END_ARRAY)) {
+        while (!JsonToken.END_ARRAY.equals(p.currentToken())) {
             builder.add(p.getValueAsInt());
             p.nextToken();
         }
--- a/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
+++ b/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
@@ -1,6 +1,10 @@
 package com.fasterxml.jackson.jr.ob.impl;
 
-import static com.fasterxml.jackson.jr.ob.impl.ValueWriterLocator.*;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.JsonTokenId;
+import com.fasterxml.jackson.jr.ob.JSONObjectException;
+import com.fasterxml.jackson.jr.ob.api.ValueReader;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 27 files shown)


jr-objects/src/test/java/com/fasterxml/jackson/jr/failing/ReadIntArray7Test.java
```diff
@@ -1,15 +1,13 @@
...
-
 import com.fasterxml.jackson.jr.ob.JSON;
 import com.fasterxml.jackson.jr.ob.TestBase;
 
-public class ReadIntArray7Test extends TestBase
-{
-    public void testReadIntArray() throws Exception
-    {
-        final int[] input = new int[] { 1, 2, 3 };
+import static org.junit.Assert.assertArrayEquals;
```
Shounaks, line 6: Move to passing.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
```diff
@@ -1,20 +1,23 @@
 package com.fasterxml.jackson.jr.ob.impl;
 
-import static com.fasterxml.jackson.jr.ob.impl.ValueWriterLocator.*;
+import com.fasterxml.jackson.core.JsonParser;
```
cowtowncoder, line 3: As per https://github.com/FasterXML/jackson/blob/master/contribution/jackson-coding-style-guide.md

(written fairly recently, to explain logic of things like import order)

import order needs to start with most general (JDK) and then to more specific (3rd party deps, core Jackson deps, finally project itself).

So let's try not to reorder these (I know IDEs like reordering things when making changes).

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
```diff
@@ -290,10 +293,17 @@ protected Path _readPath(JsonParser p) throws IOException {
         }
     }
 
-    protected int[] _readIntArray(JsonParser p) throws IOException
-    {
-        // !!! TODO
-        throw new JSONObjectException("Reading of int[] not yet implemented");
+    protected int[] _readIntArray(JsonParser p) throws IOException {
+        if (p.getCurrentToken().equals(JsonToken.START_ARRAY)) {
```
cowtowncoder, line 297: `getCurrentToken()` is deprecated, let's use `currentToken()`.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
```diff
@@ -290,10 +293,17 @@ protected Path _readPath(JsonParser p) throws IOException {
...
+            p.nextToken();
+        }
+
+        final IntStream.Builder builder = IntStream.builder();
+        while (!p.getCurrentToken().equals(JsonToken.END_ARRAY)) {
+            builder.add(p.getValueAsInt());
```
cowtowncoder, line 303: I wonder what this does for `START_OBJECT` or `START_ARRAY`? We should probably do some other validation.
I'll have to think about this.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
cowtowncoder: np. Sorry, I use Eclipse for OSS work. But if someone else has formatter I'd be happy to add to (ask others to add) formatter on that page.

@JooHyukKim do you happen to have a formatter to share?

cowtowncoder:
Ok, changed decoding logic to avoid StackOverflowException, made it bit stricter.
I think the failing test is wrong in that it has "one too many levels"; where `int` is expected we should not accept Arrays, Objects etc.

cowtowncoder (approved):

Issue #51, opened 2017-04-17
Duplicate key detection does not work for (simple) Trees


cowtowncoder:
(note: continuation from #50)

So: enabling `JSON.Feature.FAIL_ON_DUPLICATE_MAP_KEYS` does not make tree builder enable duplicate checking.

cowtowncoder:
Alas, this can not be currently implemented because `JacksonJrsTreeCodec` does not have access to configuration information -- its `readTree()` method is only passed `JsonParser` but nothing else.

Not sure what would be the best way to tackle this: one option would be to require codec to be configured separately, but that does not seem optimal.

PR #127, opened 2024-02-28
Adding Duplicate key check in JacksonJrsTreeCodec

Shounaks:
Issue #51 : A check was missing for duplicate keys.


jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
```diff
@@ -60,6 +62,9 @@ private JrsValue nodeFrom(JsonParser p) throws IOException
                 Map<String, JrsValue> values = _map();
                 while (p.nextToken() != JsonToken.END_OBJECT) {
                     final String currentName = p.currentName();
+                    if(JSON.Feature.FAIL_ON_DUPLICATE_MAP_KEYS.enabledByDefault() && values.containsKey(currentName)){
```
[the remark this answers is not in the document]
Shounaks: thanks for pointing that out. As you said, TreeCodec doesn't have JSON as a parameter, so we cannot bring the JSON value of the current state up to this point. and its not a part of this library. (and also it doesn't make sense to change a library API just for a single line of code in another API. How should we tackle this problem... ☹️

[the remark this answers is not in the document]
cowtowncoder: Ideally configuration would be passed -- maybe there is something I overlooked earlier, as I did not spend tons of time on it? I am open to suggestions, as always.

Shounaks:
@cowtowncoder , how about the extension storing the config information? That makes sense, that when we register the extension, we should have the configuration stored as well 🙌

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
```diff
@@ -32,49 +37,53 @@ public <T extends TreeNode> T readTree(JsonParser p) throws IOException {
...
+            case JsonTokenId.ID_START_ARRAY: {
                 List<JrsValue> values = _list();
                 while (p.nextToken() != JsonToken.END_ARRAY) {
                     values.add(nodeFrom(p));
                 }
                 return new JrsArray(values);
             }
-        case JsonTokenId.ID_START_OBJECT:
-            {
+            case JsonTokenId.ID_START_OBJECT: {
```
cowtowncoder, line 61: These style changes make it very difficult to review actual changes. Any chance they could be avoided?

jr-stree/src/test/java/com/fasterxml/jackson/jr/stree/JacksonJrTreeTestBase.java
```diff
@@ -44,11 +44,19 @@ protected String a2q(String json) {
         return json.replace("'", "\"");
     }
 
+    protected JSON jsonWithTreeCodec(JSON config) {
+        return JSON.builder()
+                // 13-Feb-2020, tatu: There are 2 different ways actually..
+//            .treeCodec(new JacksonJrsTreeCodec())
+                .register(new JrSimpleTreeExtension(config))
```
cowtowncoder, line 51: This is backwards incompatible change, breaking all existing users. I don't think that is doable.

cowtowncoder:
Ok, no, let's not force passing of `JSON` on constructor of codec.

Instead, we should probably add a callback in codec that `JSON` can call to pass configuration, called when codec is being registered.
That callback could pass `JSON`, or -- if there's a clean way -- pass just feature flags?

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
Shounaks: understood, will avoid this!

cowtowncoder:
@Shounaks I went ahead and added #131 so extensions now have a way of figuring out configuration. I also updated this PR to use the setting.

Shounaks:
Oh awesome! this is nice way to handle the state. I didn't know there was an Extension context!! that makes much more sense. 

Thanks for help 😃 @cowtowncoder

cowtowncoder:
Thank you @Shounaks  -- made minor tweaks, merged in for 2.17.0

PR #128, opened 2024-03-01
Adding Test Module with Groovy and Java 21 Support

Shounaks:


jackson-jr-test-module/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+                <artifactId>moditect-maven-plugin</artifactId>
+            </plugin>
+            <!-- 05-Jul-2020, tatu: Add generation of Gradle Module Metadata -->
+            <plugin>
+                <groupId>de.jjohannes</groupId>
+                <artifactId>gradle-module-metadata-maven-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+</project>
```
pjfanning, line 112: Could you ensure line endings at end of every file?

Shounaks: I tried to fix the formatting of this file, but i don't see any problem/change in it, if you are using a formatter, can you share with me, so that this can be avoided in future.

cowtowncoder: I think the end-of-file linefeed is sort of typical expectation, outside of other formatting aspects (some tools complaint about it).
I'll go ahead and fix this.

jackson-jr-test-module/pom.xml
```diff
@@ -0,0 +1,97 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <!-- This module was also published with a richer model, Gradle metadata,  -->
```
cowtowncoder, line 3: We can leave this but we probably don't want to publish jar.

jackson-jr-test-module/src/moditect/module-info.java
```diff
@@ -0,0 +1,9 @@
+module com.fasterxml.jackson.jr.annotationsupport {
```
cowtowncoder, line 1: Is this really needed even for test only?

cowtowncoder:
Ok name of module does not follow convention: i'll rename it as `jr-test-module` to be similar.

jackson-jr-test-module/pom.xml
cowtowncoder: Looks like there is trailing linefeed, all good.

jr-test-module/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+    <licenses>
+        <license>
+            <name>The Apache Software License, Version 2.0</name>
+            <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+    <properties>
+        <!--    To Test Latest Java Features Lets use this settings    -->
+        <maven.compiler.source>21</maven.compiler.source>
```
cowtowncoder, line 29: Should we use 17 instead, as that has Record type? Or is there some reason to limit to later JDKs?

cowtowncoder:
Ok. So, one big thing to change is to NOT make the test sub-project a publishable module -- we do not want to publish a jar/bundle since it is only used for testing the project itself. Otherwise publishing will push yet another useless module to Maven central.
This allows removal of stuff in `src/main/resources`, as well as half of `pom.xml`.

I forget what is the way to avoid building an artifact, but there is a way.

I renamed sub-module to follow naming convention but did not touch `pom.xml`.

Also not sure about JDK to require: but for whatever reason, I was able to build and test it with JDK 8...
so I don't think settings take effect yet (was expecting compile or test phase to fail, but everything passed).

jackson-jr-test-module/pom.xml
Shounaks: understood. if its not required, then i will remove it.

cowtowncoder:
@Shounaks I don't think Groovy test is actually run: it is compiled (I can make compilation fail by modifying class), but test not run.

This by

    ./mvnw clean test

from main level or

    ../mvnw clean test

from module dir. Both "pass", no tests run; only `testCompile` phase run.

Shounaks:
image
I was finally able to get the testcase compiled and running, after struggling for 3 days 😂 turns out, the fix was easier.

lets have a last look, and clean up the un-necessary items.

cowtowncoder (approved):

cowtowncoder:
Looks good; I made some additional cuts and can now merge to 2.17, then up to `master` (need to rename packages) for 3.0.

PR #130, opened 2024-03-07
Adding support for Java 17  & Groovy Records

Shounaks:
Continuing from #122 , Here is a cleaner implementation of the Java/Groovy Record Support with Tests.


jr-test-module/src/test/java/Java14RecordTest.java
```diff
@@ -0,0 +1,22 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * This test is in test module since the JDK version to be tested is higher than other, and hence supports Records.
+ */
+public class Java14RecordTest {
```
cowtowncoder, line 11: Records only officially part of 15 I think (in 14 preview feature) so let's call Java17 (first Long-Term Supported (LTS) JDK that has Records).

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/JSON.java
```diff
@@ -261,6 +261,13 @@ public enum Feature
          */
         USE_IS_GETTERS(true, true),
 
+        /**
+         * Feature that provides support for Groovy & JDK14 records, by allowing
```
cowtowncoder, line 265: "provides serialization support" (since this won't allow deserializing)

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/JSON.java
```diff
@@ -261,6 +261,13 @@ public enum Feature
...
+         * reading of "non-get-getters" in a class, (like for a field named <code>amount</code>
+         * the getter would be <code>amount()</code>)
+         * */
+        USE_FIELD_NAME_GETTERS(false,true),
```
cowtowncoder, line 269: I am trying to think of naming... this is accurate and might work. But wondering if we could think of even better name. `USE_` is right prefix. I wonder if... something like `USE_PLAIN_GETTERS` would better capture the.

Or, better yet, `USE_FIELD_MATCHING_GETTERS` ?

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanPropertyIntrospector.java
```diff
@@ -146,6 +152,17 @@ private static void _introspect(Class<?> currType, Map<String, PropBuilder> prop
                         _propFrom(props, name).withIsGetter(m);
                     }
                 }
+                else if (isFieldNameGettersEnabled) {
+                    // This will allow getters with field name as their getters, like the ones generated by Groovy
+                    // If method name matches with field name, & method return type matches with field type
+                    // only then it can be considered a direct name getter.
+                    final String decapName = name;
```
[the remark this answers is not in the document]
Shounaks: keeping this conversation open, 
is this the implementation we're expecting or I need to make some changes?

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanPropertyIntrospector.java
```diff
@@ -146,6 +153,19 @@ private static void _introspect(Class<?> currType, Map<String, PropBuilder> prop
                         _propFrom(props, name).withIsGetter(m);
                     }
                 }
+                else if (isFieldNameGettersEnabled) {
+                    // This will allow getters with field name as their getters, like the ones generated by Groovy
+                    // If method name matches with field name, & method return type matches with field type
+                    // only then it can be considered a direct name getter.
+                    final Map<String, Field> fieldNameMap =
+                            stream(currType.getDeclaredFields()).collect(toMap(Field::getName, identity()));
```
cowtowncoder, line 161: No I meant to pre-create lookup map before looping since set of Fields does not vary by method.

Shounaks: done!

Shounaks:
Okay! PR looks ready✌️

cowtowncoder:
Merged for inclusion in 2.17.0 -- thank you @Shounaks !

PR #133, opened 2024-03-10
Proposed Cleanup items for BeanPropertyIntrospector

Shounaks:
A small effort to reduce complexity.

P.S. sorry for formatting, 😂


cowtowncoder:
This will need to wait past 2.17 release, considered for 2.18.

But one quick note: if changing functionality to static methods, sub-classing no longer possible.
I think that I may have intended this to be a possible extension point, so developers could replace it with a sub-class implementation with suitable overrides.
If so, I wouldn't want to change it from regular class into containers of static methods.

cowtowncoder:
Ah. Actually... ok, no. I was wrong: there is no way to override it, regardless of my original thinking. Annotation-support package actually has `AnnotationBasedIntrospector` which ideally should be sub-class, I think.

So there is room for re-considering what to do here. After 2.17.0 release :)

Shounaks:
Can be optimized a bit more, and JavaDocs are remaining.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanPropertyIntrospector.java
```diff
@@ -1,40 +1,42 @@
...
-public class BeanPropertyIntrospector
-{
-    protected final static Prop[] NO_PROPS = new Prop[0];
-
-    private final static BeanPropertyIntrospector INSTANCE = new BeanPropertyIntrospector();
-
-    public BeanPropertyIntrospector() { }
-
-    public static BeanPropertyIntrospector instance() { return INSTANCE; }
+public final class BeanPropertyIntrospector {
```
pjfanning, line 30: probably can't make a class final in a patch - needs a minor release ideally

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanPropertyIntrospector.java
```diff
@@ -1,40 +1,42 @@
...
+    private BeanPropertyIntrospector() {
+    }
 
-    public POJODefinition pojoDefinitionForDeserialization(JSONReader r, Class<?> pojoType) {
-        return _introspectDefinition(pojoType, false, r.features());
+    public static POJODefinition pojoDefinitionForDeserialization(JSONReader r, Class<?> pojoType) {
```
pjfanning, line 34: can't change a public method to be static in a patch

Shounaks: I get your concern, this change can have unforeseen side-effects.
Currently this is just a proof of concept, to simplify and cleanup this class. 

we can take items from here individually (if not all). or not use this at all (whatever is feasible)

Issue #90, opened 2021-11-13
`USE_BIG_DECIMAL_FOR_FLOATS` feature not working when using `JSON.treeFrom()`


jvdsandt:
This test fails because the USE_BIG_DECIMAL_FOR_FLOATS feature is ignored and Double objects are stiill used.

    @Test
    public void testParseDecimals() throws Exception {

        JSON json = JSON.builder()
                .treeCodec(new JacksonJrsTreeCodec())
                .enable(JSON.Feature.USE_BIG_DECIMAL_FOR_FLOATS)
                .build();

        String input = "[1.1]";

        TreeNode node = json.treeFrom(input);
        TreeNode elemNode = node.get(0);

        assertTrue(elemNode instanceof JrsNumber);
        assertTrue(((JrsNumber) elemNode).getValue() instanceof BigDecimal);
    }

cowtowncoder:
Thank you for reporting this. I added a failing test to reproduce the problem.

Unfortunately `JacksonJrsTreeCodec` has no access to feature flags of `JSON`, so it cannot really determine this setting at this point. To solve the problem we'd probably have to do one of:

1. Figure out how to let codec access feature settings at time of registration (I am  not sure it is quite possible to have fully dynamic link for new `JSON` instances)
2. Add separate setting(s) for codec to be configured to do coercion.

cowtowncoder:
Fixed for 2.17.1.

NOTE: **MUST** register codec as extension like so:

```
        JSON json = JSON.builder()
                .enable(JSON.Feature.USE_BIG_DECIMAL_FOR_FLOATS)
                .register(new JrSimpleTreeExtension())
                .build();
```

for things to work; registering directly as `TreeCodec` will not work.

PR #135, opened 2024-03-16
Fix: Use Big Decimal for Tree

Shounaks:
For Issue: #90


jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
```diff
@@ -50,7 +51,16 @@ private JrsValue nodeFrom(JsonParser p) throws IOException
             return JrsBoolean.FALSE;
         case JsonTokenId.ID_NUMBER_INT:
         case JsonTokenId.ID_NUMBER_FLOAT:
-            return new JrsNumber(p.getNumberValue());
+            if (_useBigDouble) {
```
pjfanning, line 54: what does bigDouble mean?

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
```diff
@@ -28,10 +28,11 @@ public JacksonJrsTreeCodec(ObjectCodec codec) {
     }
 
     // @since 2.17
-    public void setFailOnDuplicateKeys(boolean state) {
-        _failOnDuplicateKeys = state;
+    public void setFailOnDuplicateKeys(boolean failOnDuplicateKeys,boolean useBigDouble) {
```
pjfanning, line 31: you cant change a public API in a patch? revert this. You can add extra methods but you cant change the signature of a public method without deprecating it at least a minor release.

pjfanning (changes requested):

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
Shounaks: Removed!

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
Shounaks: changed the name to be more understandable

Shounaks:
I believe this is the best route of action we can change without changing the base API for `TreeCodec`, only if that had the ability of storing the features, or atleast passing the features throught `TreeCodec.readTree(...)` then it would be much easier.

cowtowncoder:
Quick note: needs to go against `2.18` branch.

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
```diff
@@ -50,6 +65,16 @@ private JrsValue nodeFrom(JsonParser p) throws IOException
             return JrsBoolean.FALSE;
         case JsonTokenId.ID_NUMBER_INT:
         case JsonTokenId.ID_NUMBER_FLOAT:
+            final JsonParser.NumberType n = p.getNumberType();
```
cowtowncoder, line 68: Cannot use this method, have a look at jackson-databind `JsonNodeDeserializer` (I think) for better version. Problem being that this method will coerce type because JSON has no native knowledge of optimal FP type to use.

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
```diff
@@ -50,6 +65,16 @@ private JrsValue nodeFrom(JsonParser p) throws IOException
...
+            if (n == JsonParser.NumberType.FLOAT) {
+                return new JrsNumber(p.getFloatValue());
+            }
+            if (_useBigDecimalForFloat && n == JsonParser.NumberType.DOUBLE) {
```
cowtowncoder, line 72: second check unnecessary; if flag is set, `FLOAT` would also be coerced not just `DOUBLE`.

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
```diff
@@ -13,11 +14,25 @@
...
 {
-    public static JrsMissing MISSING = JrsMissing.instance;
+    public static final JrsMissing MISSING = JrsMissing.instance;
     protected final ObjectCodec _objectCodec;
 
     // @since 2.17
     protected boolean _failOnDuplicateKeys;
+    protected boolean _useBigDecimalForFloat;
+
+    public JacksonJrsTreeCodec withFeature(JSON.Feature... features) {
```
cowtowncoder, line 24: I don't think it's a good idea to require separate configurability of tree codec when `JSON` already has settings. There needs to be a way for codec to access those settings.

So why wouldn't this use the approach already added in #51?

jr-stree/src/test/java/com/fasterxml/jackson/jr/stree/failing/ReadAsBigDecimal90Test.java
```diff
@@ -13,8 +13,7 @@ public class ReadAsBigDecimal90Test extends JacksonJrTreeTestBase
     public void testReadAsBigDecimal() throws Exception
     {
         JSON json = JSON.builder()
-                .treeCodec(new JacksonJrsTreeCodec())
-                .enable(JSON.Feature.USE_BIG_DECIMAL_FOR_FLOATS)
+                .treeCodec(new JacksonJrsTreeCodec().withFeature(JSON.Feature.USE_BIG_DECIMAL_FOR_FLOATS))
```
cowtowncoder, line 16: No, the existing code is the proper way to do it (except for `enable` should probably be added before `treeCodec()` call.

cowtowncoder:
I am confused: #51 already showed the way to pass JSON.Feature flags... why wouldn't that approached be used?

Also: care must be taken when checking expected FP type; there have been lots of fine-tuning in `jackson-databind` to avoid changing natural type of non-JSON formats.

Shounaks:
Hi @cowtowncoder , I know about the other PR, but was hesitant to use Extension instead of already present constructs. Yes this can be done as you say (and i have made the changes for the same).

jr-stree/src/test/java/com/fasterxml/jackson/jr/stree/failing/ReadAsBigDecimal90Test.java
Shounaks: done

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
Shounaks: My Initial concerns were that we need to use extensions, instead of already present classes.

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
Shounaks: done!

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
cowtowncoder: Tree codec is already added as an extension isn't it.

cowtowncoder (approved):
LGTM -- will merge later today

cowtowncoder:
@Shounaks Ahh! I finally understood what you meant -- if using `JSON.builder().treeCodec()` this wouldn't work. Good point.

Still, I think I will create a separate issue for deprecating that method so that tree codecs will rather be registered as extensions and not directly assigned as `TreeCodec`s -- exactly because of this problem.

cowtowncoder (approved):

Shounaks:
yes exactly!
>  I will create a separate issue for deprecating that method

Thanks!

jr-stree/src/main/java/com/fasterxml/jackson/jr/stree/JacksonJrsTreeCodec.java
```diff
@@ -50,6 +59,9 @@ private JrsValue nodeFrom(JsonParser p) throws IOException
             return JrsBoolean.FALSE;
         case JsonTokenId.ID_NUMBER_INT:
         case JsonTokenId.ID_NUMBER_FLOAT:
+            if (_useBigDecimalForDouble) {
```
cowtowncoder, line 62: Actually, this has bug -- now all integer numbers become `BigDecimal`s too. I'll fix it.

PR #148, opened 2024-06-11
Add support for Java Record deserialization

TomaszGaweda:
Added support for record deserialization.

It is based on new Java's reflection methods, added with records. If record classes are not supported, code should follow normal path.

Difference from standard deserialization: properties are not set one-by-one, but caches in `values` array and then passed to all-arg constructor (always present in records)

### DIFF
--- a/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanConstructors.java
+++ b/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanConstructors.java
@@ -13,6 +13,8 @@ public class BeanConstructors
     protected final Class<?> _valueType;
 
     protected Constructor<?> _noArgsCtor;
+
+    // @since 2.18
     protected Constructor<?> _recordCtor;
 
     protected Constructor<?> _intCtor;
@@ -28,6 +30,7 @@ public class BeanConstructors
         return this;
     }
 
+    // @since 2.18
     public BeanConstructors addRecordConstructor(Constructor<?> ctor) {
         _recordCtor = ctor;
         return this;
@@ -73,7 +76,8 @@ public class BeanConstructors
         return _noArgsCtor.newInstance((Object[]) null);
     }
 
-    protected Object create(Object[] components) throws Exception {
+    // @since 2.18
+    protected Object createRecord(Object[] components) throws Exception {
         if (_recordCtor == null) {
             throw new IllegalStateException("Class "+_valueType.getName()+" does not have record constructor to use");
         }
--- a/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanConstructors.java
+++ b/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanConstructors.java
@@ -13,6 +13,7 @@ public class BeanConstructors
     protected final Class<?> _valueType;
 
     protected Constructor<?> _noArgsCtor;
+    protected Constructor<?> _recordCtor;
 
     protected Constructor<?> _intCtor;
     protected Constructor<?> _longCtor;
@@ -24,6 +25,11 @@ public class BeanConstructors
 
     public BeanConstructors addNoArgsConstructor(Constructor<?> ctor) {
         _noArgsCtor = ctor;
+        return this;
+    }
+
+    public BeanConstructors addRecordConstructor(Constructor<?> ctor) {
+        _recordCtor = ctor;
         return this;
     }
 
@@ -46,6 +52,9 @@ public class BeanConstructors
         if (_noArgsCtor != null) {
             _noArgsCtor.setAccessible(true);
         }
+        if (_recordCtor != null) {
+            _recordCtor.setAccessible(true);
+        }
         if (_intCtor != null) {
             _intCtor.setAccessible(true);
         }
@@ -62,6 +71,13 @@ public class BeanConstructors
             throw new IllegalStateException("Class "+_valueType.getName()+" does not have default constructor to use");
         }
         return _noArgsCtor.newInstance((Object[]) null);
+    }
+
+    protected Object create(Object[] components) throws Exception {
+        if (_recordCtor == null) {
+            throw new IllegalStateException("Class "+_valueType.getName()+" does not have record constructor to use");
+        }
+        return _recordCtor.newInstance(components);
     }
     
     protected Object create(String str) throws Exception {
--- a/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
+++ b/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
@@ -155,7 +155,7 @@ public class BeanReader
                 return _constructors.create(p.getLongValue());
             case START_OBJECT:
                 {
-                    if (RecordsHelpers.isRecord(_valueType)) {
+                    if (RecordsHelpers.isRecordType(_valueType)) {
                         final List<Object> values = new ArrayList<>();
 
                         String propName;
--- a/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
+++ b/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
@@ -155,6 +155,20 @@ public class BeanReader
                 return _constructors.create(p.getLongValue());
             case START_OBJECT:
                 {
+                    if (RecordsHelpers.isRecord(_valueType)) {
+                        final List<Object> values = new ArrayList<>();
+
+                        String propName;
+                        for (; (propName = p.nextFieldName()) != null;) {
+                            BeanPropertyReader prop = findProperty(propName);
+                            if (prop == null) {
+                                handleUnknown(r, p, propName);
+                                continue;
+                            }
+                            values.add(prop.getReader().readNext(r, p));
+                        }
+                        return _constructors.create(values.toArray());
+                    }
                     Object bean = _constructors.create();
                     String propName;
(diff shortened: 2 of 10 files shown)


jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/RecordsHelpers.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+public final class RecordsHelpers {
+    private static boolean supportsRecords;
+
+    private static Method isRecordMethod;
+    private static Method getRecordComponentsMethod;
+    private static Method getTypeMethod;
+
+    static {
+        Method isRecordMethod;
```
cowtowncoder, line 21: I think we can avoid this one... see note below.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/RecordsHelpers.java
```diff
@@ -0,0 +1,85 @@
...
+                    return false;
+                }
+            }
+        } catch (IllegalAccessException | InvocationTargetException e) {
+            return false;
+        }
+        return true;
+    }
+
+    public static boolean isRecord(Class<?> clazz) {
```
cowtowncoder, line 74: Instead of this, Jackson-databind's `ClassUtil` has:

```
    public static boolean isRecordType(Class<?> cls) {
        Class<?> parent = cls.getSuperclass();
        return (parent != null) && "java.lang.Record".equals(parent.getName());
    }
```

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
```diff
@@ -155,6 +155,20 @@ public Object read(JSONReader r, JsonParser p) throws IOException
                 return _constructors.create(p.getLongValue());
             case START_OBJECT:
                 {
+                    if (RecordsHelpers.isRecord(_valueType)) {
```
cowtowncoder, line 158: To avoid doing this check on every read, could (and should) have a `boolean _isRecord` initialized in constructor.

pom.xml
```diff
@@ -28,6 +28,7 @@
     <module>jr-extension-javatime</module>
     <module>jr-test-module</module>
     <module>jr-all</module>
+      <module>jr-records</module>
```
cowtowncoder, line 31: This does not seem to be needed (can remove?)

cowtowncoder:
@TomaszGaweda GREAT contribution, thank you! I added some notes on changes needed; nothing major, just some streamlining.

One thing I'd need, aside from doing review, is CLA, from:

(unless you have done one before; apologies if I missed it)

This is only needed once before merging the first contribution. Usually it's easiest to print, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
Once I get it, I can merge the PR (pending code review etc)

Once again, big thank you for contribution this & let's get it merged in for 2.18(.0)!

pom.xml
cowtowncoder: Removed as that failed CI.

TomaszGaweda:
BTW. Should I somehow notice that RecordHelpers is based on Hazelcast's JavaRecordSerializer? I am Hazelcast's contractor, but this work is done privately. Hz code is under Apache 2 as well

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
TomaszGaweda: eb6521bef69dfe29b947d224ed8aa227c6af071f

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/RecordsHelpers.java
TomaszGaweda: 53436e6d0df3d4148186458f03fa4e28436b7350

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/RecordsHelpers.java
TomaszGaweda: 53436e6d0df3d4148186458f03fa4e28436b7350

TomaszGaweda:
CLA signed and sent, CR comments addressed :)

cowtowncoder:
I think `RecordHelpers` is general enough -- code in jackson-databind has detection code that is quite similar -- and used by many projects, so unless it's a clear-cut copy I probably would not add a reference there.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanConstructors.java
```diff
@@ -63,6 +72,13 @@ protected Object create() throws Exception {
         }
         return _noArgsCtor.newInstance((Object[]) null);
     }
+
+    protected Object create(Object[] components) throws Exception {
```
cowtowncoder, line 76: I think I'll rename this as "createRecord()" since that's what it does....

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/RecordsHelpers.java
```diff
@@ -0,0 +1,73 @@
...
+
+import com.fasterxml.jackson.jr.ob.impl.POJODefinition.PropBuilder;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Map;
+
+/**
+ * Helper class to get Java Record metadata.
```
cowtowncoder, line 11: @since 2.18

cowtowncoder:
Thank you @TomaszGaweda -- made some final tweaks (wrt since tags, add release notes), ready to go!

In future there may need to try to support annotation override (via jackson annotations extension), but for basic usage this should work nicely..

cowtowncoder:
Hmmh. Unfortunately, didn't realize this but I don't think our Record test is running at all... 

Will need to figure why, how to fix -- it appears to try to use JUnit 5, although all (other) tests are JUnit4

jr-test-module/src/test/java/Java17RecordTest.java
```diff
@@ -12,9 +13,15 @@ public class Java17RecordTest {
...
     public void testJava14RecordSupport() throws IOException {
+        JSON jsonParser = JSON.builder().enable(Feature.USE_FIELD_MATCHING_GETTERS).build();
         var expectedString = "{\"message\":\"MOO\",\"object\":{\"Foo\":\"Bar\"}}";
-        var json = JSON.builder().enable(JSON.Feature.USE_FIELD_MATCHING_GETTERS).build().asString(new Cow("MOO", Map.of("Foo", "Bar")));
+        Cow expectedObject = new Cow("MOO", Map.of("Foo", "Bar"));
+
+        var json = jsonParser.asString(expectedObject);
         Assert.assertEquals(expectedString, json);
+
+        Cow object = jsonParser.beanFrom(Cow.class, json);
```
darxriggs, line 23: Here is some feedback on other variants of input data.

a) different order of fields in JSON than in Java record
input: `{"object":{"Foo":"Bar"}, "message":"MOO"}`
result: `Failed to create an instance of jr.Java17RecordTest$Cow due to (java.lang.IllegalArgumentException): argument type mismatch`

This should get fixed.

b) missing field in JSON
input: `{"message":"MOO"}`
result: `Failed to create an instance of jr.Java17RecordTest$Cow due to (java.lang.IllegalArgumentException): wrong number of arguments`

I am wondering how deserialization to records should handle missing data as the all-args record constructor requires all fields and if there should also be support for `Optional` like provided by `jackson-modules-java8`.

cowtowncoder: @darxriggs Could you file a separate issue for the problem with ordering? We definitely need more testing here. It can include part about missing values as well (no need for 2 issues).
As to missing values: should pass `null`, although this will be problematic for many cases like primitives.
These can be addressed incrementally; `Optional` isn't yet supported for anything else yet so it can be deferred.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
```diff
@@ -155,6 +158,22 @@ public Object read(JSONReader r, JsonParser p) throws IOException
...
+                        final List<Object> values = new ArrayList<>();
+
+                        String propName;
+                        for (; (propName = p.nextFieldName()) != null;) {
+                            BeanPropertyReader prop = findProperty(propName);
+                            if (prop == null) {
+                                handleUnknown(r, p, propName);
+                                continue;
+                            }
+                            values.add(prop.getReader().readNext(r, p));
```
cowtowncoder, line 172: This is wrong: it cannot just blindly append entries -- needs to figure which index it should go to.

I should have caught this in code review.

cowtowncoder:
Ok, no, this simply does not work as-is. I will have to revert this for now.

jr-test-module/src/test/java/Java17RecordTest.java
TomaszGaweda: Thanks @darxriggs, I've included your feedback in a follow-up: https://github.com/FasterXML/jackson-jr/pull/163

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
TomaszGaweda: Fixed in https://github.com/FasterXML/jackson-jr/pull/163

Issue #154, opened 2024-06-14
JDK17 Record serialization should work with default settings (no need for `JSON.Feature.USE_FIELD_MATCHING_GETTERS`)


cowtowncoder:
(off-shoot of #149 )

While #94 added support for serializing Java (17) and Groovy records, it requires explicit enabling of `JSON.Feature.USE_FIELD_MATCHING_GETTERS` (as per tests that now run).
But this should not be needed; code should automatically apply this setting for Record types.

Note that #148 adds introspection capability to detect if given class is a Record.

cowtowncoder:
@Shounaks do you think you could help resolve this?

cowtowncoder:
Fixed via #156

PR #156, opened 2024-06-14
Allowing Records to bypass FieldNameGetterCheck

Shounaks:
#154 : Letting Records bypass the check for USE_FIELD_MATCHING_GETTERS.


cowtowncoder:
This would make sense, but unfortunately tests now fail: did you try running them?

Shounaks:
yes, these are running on local, but I dont understand why they are failing the build...

cowtowncoder:
> yes, these are running on local, but I don't understand why they are failing the build... The changes doesn't look like they should break anything.

Yes, agreed. Ok, I'll try to see: things do fail when run locally for me.

cowtowncoder:
@Shounaks it is actually quite logical when looking at NPE... fixed.

Shounaks:
> @Shounaks it is actually quite logical when looking at NPE... fixed.

That's more strange, then why was the build failing in only one jdk version? 😆  
image
I don't understand this behavior...

cowtowncoder:
That's easily explained: because Records are not available for JDK 8 or 11, I had to make test only available for JDK17 Maven profile. And since I did not realize profile as defined ONLY runs if building with JDK 17, JDK 21 did not run tests either.
If this was not done, JDK 8 and JDK 11 test runs would have failed on compilation.

But I fixed that problem so 17 and 21 failed (8 and 11 passed because no tests were included).

You can check out `jr-record-test/pom.xml` to see how this works.

Shounaks:
Ohh ok, so that was causing my Java17 setup to not fail in local because I was running all the tests together, rather than running this individual test, and for you it was failing on local most probably due to the same JDK version, hence the mismatch. Thanks for the explanation and guidance.

cowtowncoder:
Well, actually, there was a problem with `pom.xml` when Groovy tests and Record test were combined -- Record test never even compiled.
The problem was that the packaging type was defined as "pom", which then apparently builds nothing (or rather compiles nothing, and then can't run tests).
I had to change that to "jar" (or "bundle" would work too).

And I figured out a way to prevent publishing of artifacts: we unfortunately have useless 2.17.0 and 2.17.1 pom releases but should not publish any more of those.

So basically project setup for tests was broken and no one (myself included) realized... since it did actually run the Groovy test.

Issue #158, opened 2024-06-17
 Improve support for Java Record deserialization (absent fields)


darxriggs:
Basic support was added in #148. Its implementation assumes that all fields of a Java record are present in the JSON.

This issue is about improving the deserialization to support absent fields in the JSON.
Currently deserialization fails with `Failed to create an instance of SomeClass due to (java.lang.IllegalArgumentException): wrong number of arguments` in case the amount of fields doesn't match (see [this comment](https://github.com/FasterXML/jackson-jr/pull/148#discussion_r1641769406) on the initial implementation).

There are at least two cases that should be considered:

1) Record field of a _reference type_
This should be treated as an optional field.
Deserialization should therefore set the record field value to `null` in case this field is absent in the JSON.

2) Record field of a _primitive type_
This should be treated as a mandatory field.
Deserialization should therefore fail with a useful message in case this field is absent in the JSON.

cowtowncoder:
Initial Record deserialization support reverted due to issues, closing.

PR #159, opened 2024-06-18
Improving Support for Record Deserialization

Shounaks:
#158  - Trying to accommodate null objects in records. 
this look like the easiest fix, but not sure if there is any better way.

cc- @TomaszGaweda, @cowtowncoder

### DIFF
--- a/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
+++ b/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
@@ -2,6 +2,7 @@ package com.fasterxml.jackson.jr.ob.impl;
 
 import java.io.IOException;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
 import java.util.*;
 
 import com.fasterxml.jackson.core.JsonParser;
@@ -161,6 +162,7 @@ public class BeanReader
                     // [jackson-jr#148] Record deser support (2.18)
                     if (_isRecordType) {
                         final List<Object> values = new ArrayList<>();
+                        final List<Object> props = new ArrayList<>();
 
                         String propName;
                         for (; (propName = p.nextFieldName()) != null;) {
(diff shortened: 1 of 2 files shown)


jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
```diff
@@ -161,6 +162,7 @@ public Object read(JSONReader r, JsonParser p) throws IOException
                     // [jackson-jr#148] Record deser support (2.18)
                     if (_isRecordType) {
                         final List<Object> values = new ArrayList<>();
+                        final List<Object> props = new ArrayList<>();
```
Shounaks, line 165: only if we can store props and its value as a pair in a single list, it would've been simpler.. 🙂

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
```diff
@@ -169,8 +171,18 @@ public Object read(JSONReader r, JsonParser p) throws IOException
                                 handleUnknown(r, p, propName);
                                 continue;
                             }
+                            props.add(propName);
                             values.add(prop.getReader().readNext(r, p));
                         }
+
+                        for(Field f : _valueType.getDeclaredFields()) {
```
cowtowncoder, line 178: I don't think this is proper way to do this: we should fix the setting to find correct index, and as part of that process (or post-processing).
Or build data structures before actual read at any rate; not every time read occurs.

cowtowncoder:
Thank you @Shounaks for contributing this, but I don't think it's a good way to address that problem.
We need to first fix the assumed-ordering flaw, and then handle primitives along with that.

Shounaks:
Understandable, closing this PR. will think of how this can be solved.
>  assumed-ordering flaw
How did you fix this in jackson-databind? any leads/ideas??

cowtowncoder:
@Shounaks implementation just needs to keep track of positions and properties of constructor parameters. I had to revert actual deserialization support (#148) because it is not straight-forward to fix, missing the position and probably requiring an abstraction or two to encapsulate details, not just reference to Constructor itself.
Jackson-databind implementation is much more complicated overall, so not portable as-is.

PR #160, opened 2024-06-19
(WIP) Fix #157: handle Record parameters in any order

### LINKED ISSUE #157: Improve support for Java Record deserialization (field order)
Basic support was added in #148. Its implementation assumes the same order of fields in the Java record and JSON.

This issue is about improving the deserialization to support any order of fields in the JSON.
Currently deserialization fails with `Failed to create an instance of SomeClass due to (java.lang.IllegalArgumentException): argument type mismatch` in case the field order doesn't match (see [this comment](https://github.com/FasterXML/jackson-jr/pull/148#discussion_r1641769406) on the initial implementation).

cowtowncoder:


cowtowncoder:
Ok no easy path forward, will close this, revert #148.

PR #169, opened 2024-10-18
Solve ordering issue of record constructor parameters

### LINKED ISSUE #167: Deserialization of record fails on constructor parameter ordering
I was testing if I could use record deserialization in my project and I found that it fails to match the correct parameter in the record constructor. I seems to follow alphabetic order. I'm not sure if this can differ in other JDK implementations. I'm using `azul-21`.

The [`ValueReaderLocator#_resolveBeanForDeser()`](https://github.com/FasterXML/jackson-jr/blob/8a03efb89249c8aed2273a7eb8f89bf762a8cabf/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java#L450) gets the parameters from the record, which seem to be returned in alphabetic order. This ordering will then be used when iterating over that list and thus using the incorrect constructor index when creating the `BeanPropertyReader` [here](https://github.com/FasterXML/jackson-jr/blob/8a03efb89249c8aed2273a7eb8f89bf762a8cabf/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java#L499). If all types match it will continue to parse and will return incorrect values. Those are eventually read [in the `BeanReader`](https://github.com/FasterXML/jackson-jr/blob/8a03efb89249c8aed2273a7eb8f89bf762a8cabf/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java#L139) based on this indexing as shown in the test provided. If they don't match it results in a type mismatch exception.

I have added a test below to reproduce the issue, but you can recreate it with the `Cow` examples as well by renaming the fields.
```java
record FoundDependency(String id, String g, String a, String v, String timestamp) {}

public void testOrderOfParamsFails() throws Exception {
    final var input = """
        {
          "id": "org.apache.maven:maven-core:3.9.8",
          "g": "org.apache.maven",
          "a": "maven-core",
          "v": "3.9.8",
          "p": "jar",
          "timestamp": 1718267050000,
          "ec": [
            "-cyclonedx.json",
            "-sources.jar",
            "-cyclonedx.xml",
            ".pom",
            "-javadoc.jar",
            ".jar"
          ],
          "tags": [
            "core",
            "maven",
            "classes"
          ]
        }
        """;
    final var expected = new FoundDependency("org.apache.maven:maven-core:3.9.8", "org.apache.maven", "maven-core", "3.9.8", "1718267050000");
    final var actual = jsonHandler.beanFrom(FoundDependency.class, input);
    assertEquals(expected, actual);
}
```

Giovds:
Closes #167

### DIFF
--- a/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
+++ b/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
@@ -452,7 +452,6 @@ public class ValueReaderLocator
         final Map<String, BeanPropertyReader> propMap;
         Map<String, String> aliasMapping = null;
 
-        boolean isRecord = RecordsHelpers.isRecordType(raw);
         if (len == 0) {
             propMap = Collections.emptyMap();
         } else {
@@ -461,6 +460,7 @@ public class ValueReaderLocator
                     // 13-May-2021, tatu: Let's retain ordering here:
                     : new LinkedHashMap<>();
             final boolean useFields = JSON.Feature.USE_FIELDS.isEnabled(_features);
+            final boolean isRecord = RecordsHelpers.isRecordType(raw);
             for (int i = 0; i < len; ++i) {
                 POJODefinition.Prop rawProp = rawProps.get(i);
                 Method setter = rawProp.setter;
@@ -496,7 +496,17 @@ public class ValueReaderLocator
                     }
                 }
 
-                propMap.put(rawProp.name, new BeanPropertyReader(rawProp.name, field, setter, i));
+                if (isRecord) {
+                    Parameter[] parameters = constructors._recordCtor.getParameters();
+                    for (int j = 0; j < parameters.length; j++) {
+                        Parameter parameter = parameters[j];
+                        if (!parameter.getName().isBlank() && parameter.getName().equals(rawProp.name)) {
+                            propMap.put(rawProp.name, new BeanPropertyReader(rawProp.name, field, setter, j));
+                        }
+                    }
+                } else {
+                    propMap.put(rawProp.name, new BeanPropertyReader(rawProp.name, field, setter, i));
+                }
 
                 // 25-Jan-2020, tatu: Aliases are a bit different because we can not tie them into
                 //   specific reader instance, due to resolution of cyclic dependencies. Instead,
--- a/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
+++ b/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
@@ -458,7 +458,7 @@ public class ValueReaderLocator
             final Parameter[] parameters = constructors._recordCtor.getParameters();
             for (int i = 0; i < parameters.length; i++) {
                 Parameter parameter = parameters[i];
-                if (!parameter.getName().isBlank()) {
+                if (!parameter.getName().isEmpty()) {
                     ctorParameterIndexMap.put(parameter.getName(), i);
                 }
             }
--- a/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
+++ b/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
@@ -452,6 +452,18 @@ public class ValueReaderLocator
         final Map<String, BeanPropertyReader> propMap;
         Map<String, String> aliasMapping = null;
 
+        final Map<String, Integer> ctorParameterIndexMap = new HashMap<>();
+        final boolean isRecord = RecordsHelpers.isRecordType(raw);
+        if (isRecord) {
+            final Parameter[] parameters = constructors._recordCtor.getParameters();
+            for (int i = 0; i < parameters.length; i++) {
+                Parameter parameter = parameters[i];
+                if (!parameter.getName().isBlank()) {
+                    ctorParameterIndexMap.put(parameter.getName(), i);
+                }
+            }
+        }
+
         if (len == 0) {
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Ok thank you @Giovds ! I'll need to go over this with more thought, but on first look it looks good!

One thing I'll need before merging -- if not already done earlier -- is to get a CLA. It needs to be sent just once before first contribution and is good for any later contributions.
It's this doc:

and the usual ways is to print, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
Once I have it, I can proceed with merging (pending final code review etc).

Looking forward to merging this in!

Giovds:
I've sent the email 👍

cowtowncoder:
CLA received, hoping to review this soon.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
```diff
@@ -452,7 +452,18 @@ protected BeanReader _resolveBeanForDeser(Class<?> raw, POJODefinition beanDef)
         final Map<String, BeanPropertyReader> propMap;
         Map<String, String> aliasMapping = null;
 
-        boolean isRecord = RecordsHelpers.isRecordType(raw);
+        final Map<String, Integer> ctorParameterIndexMap = new HashMap<>();
```
cowtowncoder, line 455: Ok. While this works, technically, I think I'd like original properties to be returned in declaration order, so that index would match. I am not (yet) sure why the order gets shuffled, will try to read code further.

cowtowncoder: Oh, something probably sorts properties alphabetically which leads to discrepancy.
So that serialization order is alphabetic by default.

There's also then the question wrt Records: should they

1. Follow POJO logic of sorting by alphabetic order, OR
2. Use Record declaration order

To me (2) is probably more often default to use, but this probably should be configurable option.
And for backwards compatibility should probably default to (1) and allow (2) via configuration.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
```diff
@@ -452,7 +452,18 @@ protected BeanReader _resolveBeanForDeser(Class<?> raw, POJODefinition beanDef)
...
+        final boolean isRecord = RecordsHelpers.isRecordType(raw);
+        if (isRecord) {
+            final Parameter[] parameters = constructors._recordCtor.getParameters();
+            for (int i = 0; i < parameters.length; i++) {
+                Parameter parameter = parameters[i];
+                if (!parameter.getName().isEmpty()) {
```
cowtowncoder, line 461: Why such check? Does this happen? Isn't it an error condition?

And specifically if this happens wouldn't it just lead to an NPE when looking up the index?

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
cowtowncoder: Ok it all happens in `BeanPropertyIntrospector`... due to

    Map<String,PropBuilder> propsByName = new TreeMap<>();

changing of which is not trivial I suppose. And problem being that Record support is based on Fields and/or Methods and not just the canonical constructor.
But I think it can be hacked a bit to use declaration order if

1. Type is record AND
2. We are introspection for Deserialization

(and with new possible feature, also for Serialization -- but that'd be for 2.19)

Giovds: The record constructor does return them in declaration order, the 
```
final List<POJODefinition.Prop> rawProps = beanDef.getProperties();
``` 
(which is probably created through the `BeanPropertyIntrospector`) returns them in alphabetic order.

Changing how that internally works seems to me like something that should be proposed in a separate PR as it might impact more than just this deserialisation part and might require some rethinking of (de)serialisation of records?

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
Giovds: Actually, I think this is an irrelevant check, as you can not create a (record) instance variable without a name... It would lead to a NPE in the look-up part. I'd say just remove the check, as you can not test this? WDYT?

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
cowtowncoder: @Giovds Right -- what I may do is to take tests from here, but make code change differently.
I realized that the way Record introspection works is bit different than I'd like -- trying to be minimal wrt changes from POJOs -- but I think I can get ordering to work in a way that no sorting should be needed.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueReaderLocator.java
cowtowncoder: Yes, sounds good. But as per my note above, might be this part of code can be removed altogether.

cowtowncoder:
Ok; I decided to go with #170 instead. Thank you for your help @Giovds reporting the issue & showing a possible fix. I'll re-visit changes for 2.19 as well.

Giovds:
Happy to see it fixed. Hopefully it was of any help 👍

cowtowncoder:
@Giovds It definitely was -- thank you for helping here.

PR #179, opened 2024-12-10
Remove jdk17 test sources post `databind#4820`

JooHyukKim:


.java-version
```diff
@@ -0,0 +1 @@
+temurin64-17.0.8.1
```
cowtowncoder, line 1: I assume this is accidental addition?

JooHyukKim: Right 🥲
It's from `jenv` I used to manage local, per-project java versions.
THankss

cowtowncoder (approved):

Issue #176, opened 2024-11-24
Cannot Deserialize JSON with `@JsonProperty`-annotated Record fields


juulhobert:
I'm encountering an issue when using Jackson-jr with annotation support to handle record classes that map to JSON with snake_case field names. While serialization works as expected, deserialization of the same JSON back into the record fails.

To illustrate the problem, I created a minimal reproducer: [jackson-jr-annotations-reproducer](https://github.com/juulhobert/jackson-jr-annoations-reproducer).

**Steps to Reproduce:**
Use a record class where fields are annotated with @JsonProperty to specify the snake_case equivalent of the field name. Serialize an instance of the record class to JSON using Jackson-jr (with annotation support). Attempt to deserialize the resulting JSON back into the record class.

Expected Behavior:
The JSON should deserialize correctly into the record class, matching fields based on the @JsonProperty annotations.

Actual Behavior:
Deserialization fails with an exception
![image](https://github.com/user-attachments/assets/378c4f7e-9d8c-4b2d-9d61-34f572bc6887)

Environment:
Jackson-jr version: 2.18.1
Java version: 21

Reproducer Link:
https://github.com/juulhobert/jackson-jr-annoations-reproducer

cowtowncoder:
Right, Record-support is pretty limited and probably won't yet work with annotation support. Thank you for reporting this.

cowtowncoder:
Fixed for 2.21.0 and 3.0.1.

PR #189, opened 2025-02-14
Fix snake case deserialization

juulhobert:
Issue: #176

### DIFF
--- a/jr-annotation-support/pom.xml
+++ b/jr-annotation-support/pom.xml
@@ -76,6 +76,14 @@
         <groupId>de.jjohannes</groupId>
         <artifactId>gradle-module-metadata-maven-plugin</artifactId>
       </plugin>
+        <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <configuration>
+                <source>16</source>
+                <target>16</target>
+            </configuration>
+        </plugin>
     </plugins>
   </build>
 
--- a/jr-annotation-support/src/test/java/com/fasterxml/jackson/jr/annotationsupport/BasicAliasTest.java
+++ b/jr-annotation-support/src/test/java/com/fasterxml/jackson/jr/annotationsupport/BasicAliasTest.java
@@ -1,6 +1,7 @@
 package com.fasterxml.jackson.jr.annotationsupport;
 
 import com.fasterxml.jackson.annotation.JsonAlias;
+import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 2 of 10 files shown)


cowtowncoder:
First of all, thank you @juulhobert for working on this! I know it's a draft and may take some time.

Eventually if all goes well, we'll need CLA (assuming you haven't sent one earlier): only needs to be done once (good for all future contributions)

It's here:

and is usually printed, filled & signed, photo/scanned, emailed to `cla` at fasterxml dot com.

But this is only need to merge, just mentioning it now.

Will also have a look at changes now.

jr-annotation-support/pom.xml
```diff
@@ -76,6 +76,14 @@
         <groupId>de.jjohannes</groupId>
         <artifactId>gradle-module-metadata-maven-plugin</artifactId>
       </plugin>
+        <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <configuration>
+                <source>16</source>
```
cowtowncoder, line 83: I assume this is just for testing -- Jackson 2.x is JDK 8, still.

jr-annotation-support/src/test/java/com/fasterxml/jackson/jr/annotationsupport/BasicAliasTest.java
```diff
@@ -34,6 +35,11 @@ public String getMiddle() {
         public void setLast(String str) { last = str; }
     }
 
+    record SnakeCaseRecord(
```
cowtowncoder, line 38: Cannot be done here, as it requires JDK beyond 8. Any Record tests need to go in separate `jr-record-test` Maven sub-project.

jr-annotation-support/pom.xml
juulhobert: Thank you for taking a look at this draft PR. I appreciate the time and effort you invest in this open-source project :) The code in this PR still needs some refinement. I already expected that changing the JDK version is not wishful. This is just the first draft, where I focused on getting the functionality to work and gaining a better understanding of the library's internals.

Can I ask you a question about the RecordsHelper class? I've changed the visibility from default to public to use it in the AnnotationBasedIntrosepector class. I'm not sure if this change is appropriate. What do you think?

cowtowncoder: @juulhobert Thanks! Apologies for slow follow up (was on vacation for a week).

So: yes, changing it to `public` is ok if (when) it is necessary.

PR #197, opened 2025-06-17
Add support for reading `boolean[]`/`short[]`/`float[]` values

lukehutch:
Proposed fix for https://github.com/FasterXML/jackson-jr/issues/210

Created with Claude Code, so should be double-checked :-)

### DIFF
--- a/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/JSONWriter.java
+++ b/jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/JSONWriter.java
@@ -191,6 +191,15 @@ public class JSONWriter
         case SER_BOOLEAN_ARRAY:
             writeBooleanArrayField(fieldName, (boolean[]) value);
             return;
+        case SER_SHORT_ARRAY:
+            writeShortArrayField(fieldName, (short[]) value);
+            return;
+        case SER_FLOAT_ARRAY:
+            writeFloatArrayField(fieldName, (float[]) value);
+            return;
+        case SER_DOUBLE_ARRAY:
+            writeDoubleArrayField(fieldName, (double[]) value);
+            return;
         case SER_TREE_NODE:
             writeTreeNodeField(fieldName, (TreeNode) value);
             return;
@@ -318,6 +327,15 @@ public class JSONWriter
             return;
         case SER_BOOLEAN_ARRAY:
             writeBooleanArrayValue((boolean[]) value);
+            return;
+        case SER_SHORT_ARRAY:
+            writeShortArrayValue((short[]) value);
+            return;
+        case SER_FLOAT_ARRAY:
+            writeFloatArrayValue((float[]) value);
+            return;
+        case SER_DOUBLE_ARRAY:
+            writeDoubleArrayValue((double[]) value);
             return;
         case SER_TREE_NODE:
             writeTreeNodeValue((TreeNode) value);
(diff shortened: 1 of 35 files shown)


jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueLocatorBase.java
```diff
@@ -61,6 +61,9 @@ public abstract class ValueLocatorBase
     public final static int SER_INT_ARRAY = 5;
     public final static int SER_LONG_ARRAY = 6;
     public final static int SER_BOOLEAN_ARRAY = 7;
+    public final static int SER_SHORT_ARRAY = 38;
```
cowtowncoder: Needs to be consecutive.

cowtowncoder:
@lukehutch Did you review it yourself? Or run tests?  (although TBH Sonatype Snapshot repo has issues so perhaps it is difficult). Looks like unit tests fail.

Overall makes sense, GenAI coding agents are pretty good at following patterns. But there are some aspects that need changing; adding comments.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
```diff
@@ -328,6 +343,190 @@ protected int[] _readIntArray(JsonParser p) throws IOException {
         return builder.build().toArray();
     }
 
+    protected long[] _readLongArray(JsonParser p) throws IOException {
+        if (JsonToken.START_ARRAY.equals(p.currentToken())) {
+            p.nextToken();
+        }
+
+        java.util.List<Long> values = new java.util.ArrayList<>();
```
cowtowncoder: Very inefficient, no. Have a look at how `int[]` is handled.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
```diff
@@ -328,6 +343,190 @@ protected int[] _readIntArray(JsonParser p) throws IOException {
...
+            switch (t) {
+            case JsonTokenId.ID_NUMBER_FLOAT:
+            case JsonTokenId.ID_NUMBER_INT:
+            case JsonTokenId.ID_NULL:
+                values.add(p.getValueAsLong());
+                break;
+            case JsonTokenId.ID_END_ARRAY:
+                break main_loop;
+            default:
+                throw new JSONObjectException("Failed to bind `long` element if `long[]` from value: "+
```
cowtowncoder: Funny. Copied typo from `int[]` case :)

("if" -> "of")

(will fix separately)

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/SimpleValueReader.java
```diff
@@ -328,6 +343,190 @@ protected int[] _readIntArray(JsonParser p) throws IOException {
...
+        }
+        return result;
+    }
+
+    protected boolean[] _readBooleanArray(JsonParser p) throws IOException {
+        if (JsonToken.START_ARRAY.equals(p.currentToken())) {
+            p.nextToken();
+        }
+
+        java.util.List<Boolean> values = new java.util.ArrayList<>();
```
cowtowncoder: Same as above, need to build an array to avoid all the boxing, unboxing.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueLocatorBase.java
cowtowncoder: ... which gets tricky wrt whether to renumber entries or move these to later.

I think renumbering is needed, so move `SER_TREE_NODE` etc up by number of added entries needed.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/JSONWriter.java
```diff
@@ -555,6 +573,45 @@ protected void writeBooleanArrayField(String fieldName, boolean[] v) throws IOEx
...
+            _generator.writeNumber(v[i]);
+        }
+        _generator.writeEndArray();
+    }
+
+    protected void writeDoubleArrayField(String fieldName, double[] v) throws IOException {
+        _generator.writeFieldName(fieldName);
+        writeDoubleArrayValue(v);
+    }
+
```
cowtowncoder, line 611: This class looks good.

lukehutch:
I couldn't run the tests, and yes, this is going to be imperfect, but I figured I'd submit it as a starting point to save you a lot of work. Thanks!

lukehutch:
By the way you could also continue to iterate with Claude Code -- probably the issues you found could be fixed with a couple more prompts consisting of one or two sentences.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/ValueLocatorBase.java
lukehutch: Yes, I saw that, and had the same thought, but I figured you'd know which of the two options (renumbering all later constants, or moving the other arrays to the end, which is ugly) would make the most sense.

cowtowncoder:
> By the way you could also continue to iterate with Claude Code -- probably the issues you found could be fixed with a couple more prompts consisting of one or two sentences.

My experience with Wind Surfer was... not great so I am bit of GenAI skeptic.

While tools are impressive in their own right, the one and only piece of code in Jackson that I auto-generated, turned out to be the first bug reported for 2.19.0.... :-D
(yeah I should have thought through recursive handling to see the issue but it "looked ok" -- and that's what GenAI optimizes: things that appear similar to what might pass. But I digress).

lukehutch:
Windsurf is pretty bad. Claude Code is almost magical, especially for tasks like this that really just require pattern matching, as long as the codebase is not enormous (in which case you run into context window size issues, where it will forget state each time it has to compact the context).

JooHyukKim:
@lukehutch were you able to use Claude code to look at current failures? Since we are at public repository, I suppose it should be able to fetch CI failures as well?

lukehutch:
I didn't have it look at test results or anything like that. I had it identify missing support for arrays of any of the 7 primitive types. It correctly identified that arrays of char are treated as strings, and some other special case handling.

cowtowncoder:
FWTW, serialization of `boolean[]`/`short[]`/`float[]`/`double[]` added via #203 so this is just for deserialization part now.

`double[]` case uses stream build helper so could just merge that, will see how to proceed.

**EDIT**: with #208, `double[]` deserialization supported, will edit title further.

cowtowncoder:
Created #211 -- also with Claude Code! -- with optimized builders. Closing this: huge thanks @lukehutch for showing me the way!

PR #201, opened 2025-08-05
Support for `OffsetDateTime` and `ZonedDateTime` in `jackson-jr-extension-javatime`

StanB-EKZ:
Added support for OffsetDateTime and ZonedDateTime to jackson-jr-extension-javatime. 

The most striking change here is the introduction of a static `FORMATTER` field in `JavaTimeReaderWriterProvider`. This field is used by the ValueReader classes. I introduced it because ISO 8601 string representations that included a time offset could cause `LocalDateTimeValueReader` to throw a DateTimeParsingException. Apparently reading requires a slightly more forgiving approach than writing.

The static formatter can parse ISO 8601 dates, optionally containing: milliseconds, hour offsets or zone IDs.

This did require me to get rid of the `withDateTimeFormatter()` method in `JavaTimeReaderWriterProvider`. Because allowing users to change the format, could let them break ISO 8601 related behavior. If anyone has a good idea of how to reliably restore that behavior, I would be happy to.

I also added a fairly straightforward unit test to check if parsing and formatting works as expected.


cowtowncoder:
First of all: thank you for the contribution! It sounds like useful addition.

It looks like there's a test failure (possible related to what is mentioned as non-compatible change)?

cowtowncoder:
I need to review this with some more time, but one process thing we need (if not done earlier), before I can merge it (pending review etc), is to get CLA. It's here:

and the usual was is to print it, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
This only needs to be done once per contributor; CLA is good for all future contributions to any Jackson component.

Looking forward to getting this merged!

StanB-EKZ:
I will look into CLA thanks.

It looks like my own unit test failed because I did not take into account that the server obviously does not share my time zone.

StanB-EKZ:
FYI: I submitted the CLA, fixed the unit test, and made some extra improvements to be even kinder on ISO 8601 input.

cowtowncoder:
@pjfanning @JooHyukKim WDYT?

(CLA exists fwtw just need to review and evaluate change itself)

JooHyukKim:
Will review tonight, idea itself seems... yes, why not?

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/LocalDateTimeValueReader.java
```diff
@@ -1,23 +1,37 @@
...
+        
+        if (ta instanceof ZonedDateTime) {
+            // Convert a date time that unexpectedly includes a time offset or zone ID, to a proper local date time
+            return ((ZonedDateTime)ta).withZoneSameInstant(_localZoneId).toLocalDateTime();
+        }
+        if (ta instanceof LocalDateTime) {
+            return ta;
+        }
+        
+        throw new IOException("Could not create a valid DateTime instance");
```
JooHyukKim: We can have more useful information.

1. Exception message instead of just `DateTime`, can have LocalDateTime or ZonedDateTime etc...

2. If Jackson Jr allows including source text, value of `p.getText()`, we could be more informative.

JooHyukKim:
It will be good to add even a bit of documentation, like version `@since` , etc...

Otherwise change seems okay

StanB-EKZ:
> It will be good to add even a bit of documentation, like version `@since` , etc...
> 
> Otherwise change seems okay

The original code hardly had any Javadoc comments, so I sort of went with it. If you'd prefer me to write some Javadoc code for each file that is lacking, I would be happy to do so!

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/LocalDateTimeValueReader.java
StanB-EKZ: Seems like a good idea. I will make this change sometime over the next few days.

cowtowncoder:
Yeah `@since` are good to add where applicable.

My bigger concerns have to do with things like timezones, defaulting, however.
Date/time type handling is truly one of trickier part of s/w engineering on any platform, so easy to get things wrong & difficult to resolve once behavior is added, released.

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/JavaTimeReaderWriterProvider.java
```diff
@@ -1,48 +1,104 @@
...
+    /**
+     * Create a forgiving date time formatter that allows different interpretations of ISO 8601
+     * strings to be parsed.
+     * 
+     * @param includeUtcDefault Set to <code>true</code> to set UTC to be the default offset
+     *                          for non-local date times. Set to <code>false</code> when handling
+     *                          local date times.
+     * @return Formatter
+     */
+    public static DateTimeFormatter createFormatter(boolean includeUtcDefault) {
```
cowtowncoder: Is the idea of exposing public static method so that if application code wants to configure different formatter, it can start with default settings? Or what is the idea: static methods can not be overriden & built formatter instances are immutable.

So I guess I am not sure of need/benefit of `public` method here.

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/JavaTimeReaderWriterProvider.java
```diff
@@ -1,48 +1,104 @@
...
-    private DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ISO_LOCAL_DATE_TIME;
-
-    public JavaTimeReaderWriterProvider() { }
+    private ZoneId _fallbackLocalZoneId;
+    
+    protected static final DateTimeFormatter FORMATTER = createFormatter(true);
+    protected static final DateTimeFormatter LOCAL_FORMATTER = createFormatter(false);
+    
+    public JavaTimeReaderWriterProvider() {
+        _fallbackLocalZoneId = ZoneId.systemDefault();
```
cowtowncoder: This seems like it is different from standard Jackson defaulting which for time zones/offsets tries to use global "neutral" default (UTC where possible, GMT otherwise), to avoid relying on local defaults that are arbitrary.

But I am not sure such choice exists for `ZoneId`?

StanB-EKZ: I understand what you mean. I personally think the default for a `LocalDateTime` object should probably always be in the local time for the user. Which this probably comes closest to. But it wouldn't hurt setting it to UTC to be safe, considering the user can change it. 

I have been doubting another time zone related segment myself, which is the `includeUtcDefault` boolean that occurs inside of `JavaTimeReaderWriterProvider`. Because that segment defaults `ZonedDateTime` and `OffsetDateTimes` to UTC when no time zone is included in its ISO 8601 string. I was thinking that maybe it would make more sense to default to the time zone that the user configured. But I am sort of undecided on that.

StanB-EKZ:
I found out that my code (and the original too) wasn't very kind on null values so I went ahead and changed that while I was adding Javadoc documentation.

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/JavaTimeReaderWriterProvider.java
StanB-EKZ: That was the idea, but I guess it is debatable if such a thing would ever happen. I was just future-proofing here. But it wouldn't be a big chore to change this around if ever needed, so I guess it might as well be private.

Below I mentioned that I have some doubts over the `includeUtcDefault ` parameter I introduced. If I were to change that around as described, I think the static method may need to be changed to a normal method anyway. So I would love some thoughts on that.

StanB-EKZ: Thought this over, and I suggest I'll indeed change it to default to UTC, maybe adding a convenience method to steer it to local time easily.

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/JavaTimeReaderWriterProvider.java
StanB-EKZ: Thought this over. Wikipedia says the following:

> If no UTC relation information is given with a time representation, the time is assumed to be in local time. 

So I suggest I'll change the formatter to use the user's selected time zone for dates that do not contain offset information. That will automatically mean the end of the static formatter.

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/JavaTimeReaderWriterProvider.java
cowtowncoder: @StanB-EKZ that sounds like a good idea.

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/JavaTimeReaderWriterProvider.java
cowtowncoder: Hmmmh. Ok.

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/DefaultDateTimeValueReader.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public Object read(JSONReader reader, JsonParser p) throws IOException {
+    	// SimpleValueReader allows 'Date' objects to be null, so this should probably
+    	// also be the case here.
+    	if (p.hasToken(JsonToken.VALUE_NULL)) {
+            return null;
+        }
+
```
cowtowncoder: Needs to do error handling for types other than `JsonToken.VALUE_STRING`, I think; like getting JSON Object or Array.

`BeanReader.java`, f.ex, has this as default case:

```
        throw JSONObjectException.from(p,
                "Can not create a "+_valueType.getName()+" instance out of "+_tokenDesc(p));
```

(`_tokenDesc` is a helper method in `ValueReader`)

cowtowncoder:
@StanB-EKZ looks good, sounds good. With minor touches ready to go. LMK when you think pr is complete and I can merge it in, right in time for 2.20.0 release (soon).

jr-extension-javatime/src/main/java/com/fasterxml/jackson/jr/extension/javatime/JavaTimeReaderWriterProvider.java
StanB-EKZ: The static formatters are still here. I found it to be really difficult to apply offsets automatically from the `DateTimeFormatter`. I did change the static method to be private though.

StanB-EKZ:
@cowtowncoder I made more changes than I wanted to, but I think I'm probably there now. 

I'm not super happy about the offset check that I introduced in `DefaultDateTimeValueReader`, but I could not figure out for the life of me how to do it differently and maintain ISO 8601 compliant behavior.

cowtowncoder:
@StanB-EKZ Thank you for doing all this. I hope to get this merged ASAP (bit overloaded but excited to get this in 2.20 release!)

jr-extension-javatime/src/test/java/com/fasterxml/jackson/jr/extension/javatime/LocalDateTimeReaderTest.java
```diff
@@ -0,0 +1,195 @@
...
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoField;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import com.fasterxml.jackson.jr.ob.JSON;
+import com.fasterxml.jackson.jr.ob.JSONObjectException;
+
+public class LocalDateTimeReaderTest {
```
cowtowncoder, line 17: Forgot to comment: these should have been split into different test classes (or test class renamed) -- by read/write, type.
Will rename post-merge now.

StanB-EKZ: You are right, I did notice that but completely forgot. My apologies.

PR #207, opened 2025-09-08
Fix record serialization when using JacksonAnnotationExtension

### LINKED ISSUE #176: Cannot Deserialize JSON with `@JsonProperty`-annotated Record fields

TomaszGaweda:
This PR fixes record serialization and deserialization when using Java Records with JacksonAnnotationExtension

Fixes #176

### DIFF
--- a/jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedIntrospector.java
+++ b/jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedIntrospector.java
@@ -63,16 +63,20 @@ public class AnnotationBasedIntrospector
         _isRecord = RecordsHelpers.isRecordType(type);
 
         // May need to retain order for Record serialization too
-        if (keepPropertyOrderForRecord(features)) {
+        if (keepPropertyOrderForRecord()) {
             _props = new IndexedMap<>();
         } else {
             _props = new HashMap<>();
         }
     }
 
-    private boolean keepPropertyOrderForRecord(int features) {
-        return _isRecord && _forSerialization
-                && JSON.Feature.WRITE_RECORD_FIELDS_IN_DECLARATION_ORDER.isEnabled(features);
+    /**
+     * Property order must be kept for records when:
+     * - record is deserialized
+     * - record is serialized and feature {@link JSON.Feature#WRITE_RECORD_FIELDS_IN_DECLARATION_ORDER} is enabled.
+     */
+    private boolean keepPropertyOrderForRecord() {
+        return _isRecord && (!_forSerialization || JSON.Feature.WRITE_RECORD_FIELDS_IN_DECLARATION_ORDER.isEnabled(_features));
     }
 
     public static POJODefinition pojoDefinitionForDeserialization(JSONReader r,
@@ -169,7 +173,7 @@ public class AnnotationBasedIntrospector
         List<APropBuilder> renamed = null;
         Iterator<APropBuilder> it = _props.values().iterator();
         final boolean keepIgnored = _isRecord && !_forSerialization;
-        final boolean keepPropertyOrderForRecord = keepPropertyOrderForRecord(_features);
+        final boolean keepPropertyOrderForRecord = keepPropertyOrderForRecord();
 
         while (it.hasNext()) {
             final APropBuilder prop = it.next();
--- a/jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedIntrospector.java
+++ b/jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedIntrospector.java
@@ -13,10 +13,14 @@ import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 
 import com.fasterxml.jackson.jr.ob.JSON;
 import com.fasterxml.jackson.jr.ob.impl.BeanConstructors;
+import com.fasterxml.jackson.jr.ob.impl.BeanPropertyIntrospector;
 import com.fasterxml.jackson.jr.ob.impl.JSONReader;
 import com.fasterxml.jackson.jr.ob.impl.JSONWriter;
 import com.fasterxml.jackson.jr.ob.impl.POJODefinition;
 import com.fasterxml.jackson.jr.ob.impl.RecordsHelpers;
+
+import static com.fasterxml.jackson.jr.ob.impl.BeanPropertyIntrospector.addConstructors;
+import static com.fasterxml.jackson.jr.ob.impl.BeanPropertyIntrospector.derivePropertiesFromConstructor;
 
 /**
  *
@@ -38,6 +42,7 @@ public class AnnotationBasedIntrospector
     // // // State (collected properties, related)
 
     protected Map<String, APropBuilder> _props;
+    protected IndexedMap<String, APropBuilder> _propsAsIndexed;
 
     // // // State only for deserialization:
 
@@ -64,9 +69,11 @@ public class AnnotationBasedIntrospector
 
         // May need to retain order for Record serialization too
         if (keepPropertyOrderForRecord()) {
-            _props = new IndexedMap<>();
+            _propsAsIndexed = new IndexedMap<>();
+            _props = _propsAsIndexed;
         } else {
             _props = new HashMap<>();
+            _propsAsIndexed  = null;
         }
     }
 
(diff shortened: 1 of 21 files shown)


jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/AnnotationBasedIntrospector.java
```diff
@@ -147,21 +195,31 @@ protected POJODefinition.Prop[] _pruneProperties(boolean sortProperties)
...
-                    renamed = new LinkedList<APropBuilder>();
+            String explicitName = prop.findPrimaryExplicitName(_forSerialization);
+            if (explicitName != null) {
+                APropBuilder newProp = prop.withName(explicitName);
+                if (keepPropertyOrderForRecord) {
+                    APropBuilder orig = _props.get(explicitName);
+                    if (orig != null) {
+                        newProp = APropBuilder.merge(orig, newProp);
+                    }
+                    ((IndexedMap<String, APropBuilder>) _props)
```
TomaszGaweda: rationale: we don't want to change position, because of the renaming

TomaszGaweda:
Hi @cowtowncoder, I'd very appreciate if you could take a look at this PR (of course nothing urgent). Thanks in advance!

cowtowncoder:
@TomaszGaweda sorry for delay, hoping to review this week (having to focus on 3.0.0-rc10 first). Just thought I'll add a note to let you know this is on my radar, todo list.

TomaszGaweda:
Thanks a lot @cowtowncoder for the update and no worries :)

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/IndexedMap.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+import static java.util.stream.Collectors.toSet;
+
+/**
+ * Map used when we need to maintain insertion index.
```
cowtowncoder, line 15: Should expand a bit to indicate why JDK `LinkedHashMap` wouldn't work?

TomaszGaweda: Added here: https://github.com/FasterXML/jackson-jr/pull/207/commits/4ec9eeca94d6a566e6c5ebed98dbe76e51003b85

jr-annotation-support/src/main/java/com/fasterxml/jackson/jr/annotationsupport/IndexedMap.java
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Replaces entry at the index of {@code oldKey} with a new entry: {@code newKey, value}.
+     */
+    public void replaceAtIndexOf(K oldKey, K newKey, V value) {
+        boolean stop = false;
+        for (int index = 0; index < entries.size() && !stop; index++) {
+            Entry<K, V> kvEntry = entries.get(index);
+            if (kvEntry.getKey().equals(oldKey)) {
+                entries.set(index, new SimpleEntry<>(newKey, value));
+                stop = true;
```
cowtowncoder: Why not just use `return` (or, `break`) instead of a flag?

cowtowncoder: Actually I'll just change. :)

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
```diff
@@ -149,9 +150,40 @@ private Object readRecord(JSONReader r, JsonParser p) throws Exception {
             Object value = prop.getReader().readNext(r, p);
             values[prop.getIndex()] = value;
         }
+        for (int i = 0; i < values.length; i++) {
```
cowtowncoder: This would require a comment -- explaining that for primitive types `null`s not allowed.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
```diff
@@ -149,9 +150,40 @@ private Object readRecord(JSONReader r, JsonParser p) throws Exception {
...
+            if (values[i] == null) {
+                for (BeanPropertyReader prop : _propsByName.values()) {
```
cowtowncoder: This is rather inefficient with n*n complexity (although I guess mostly n if no `null`s from missing entries?). Would be good to have either direct lookup (ordered array/List of props by index), or...

actually better idea, I think -- for records, pre-create `Object[]` with defaults for primitive values (instead of `null`s for regular object values)
Then something like:

    final Object[] values = Arrays.copyOf(preset, preset.length);

and avoid post-scanning?

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
```diff
@@ -149,9 +150,40 @@ private Object readRecord(JSONReader r, JsonParser p) throws Exception {
...
+        }
+        if (aClass.equals(boolean.class)) {
+            return Boolean.FALSE;
+        }
+        if (aClass.equals(short.class)) {
+            return (short) 0;
+        }
+        if (aClass.equals(char.class)) {
+            return (char) 0;
+        }
```
cowtowncoder, line 202: Missing `float.class` and `double.class`?

cowtowncoder (commented):
Pretty good; we are almost there! Couple of small changes needed and we are good to go!

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
cowtowncoder: (and `byte.class`)

Added.

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
cowtowncoder: @TomaszGaweda does this make sense? If not, I can try to explain in more detail.

TomaszGaweda: Hey @cowtowncoder ! Yes, it makes sense, 2n is definetely better than n^2. I will work on this PR this week, had very limited time last week unfortunatelly

TomaszGaweda: Done: [`8b9daf9` (#207)](https://github.com/FasterXML/jackson-jr/pull/207/commits/8b9daf913e26dc658a854861df33eb15b97d7935)

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
TomaszGaweda: Done

jr-objects/src/main/java/com/fasterxml/jackson/jr/ob/impl/BeanReader.java
cowtowncoder: That was quick, thank you! Will review later today.

TomaszGaweda:
Thank you a lot @cowtowncoder for review and your help!

